PR #5, opened 2020-08-04
Switch to using underscores as consistent word separator.

nuclearsandwich:
I've been a bit haphazard in the use of dashes versus underscores as the
word separator. In general I prefer dashes (shoutout to languages
without infix operators which make multi-word identifiers unambiguous
when dash-separated) but, dashes are not recommended in chef
identifiers, cookbooks in particular. In order to to allow attribute
names to align with cookbook names and not worry about alternate
spellings let's just settle on using underscores to separate words and
I'll learn to live with it.

While I'm only applying this change to the in-development ros_buildfarm
cookbook at some point we should revisit the other cookbooks and update
the attributes.


nuclearsandwich:
@Maanasa-Kotha I'm going to merge this now to keep iterating. FYI.

PR #6, opened 2020-08-07
Jenkins Server

Maanasa-Kotha:
Started Jenkins server recipe but doesn't work yet.


kitchen.yml
```diff
@@ -16,6 +16,7 @@ suites:
     data_bags_path: "test/integration/data_bags"
     run_list:
       - recipe[ros_buildfarm::agent]
+      - recipe[ros_buildfarm::jenkins]
```
nuclearsandwich: Instead of running the jenkins and agent recpies together in the agent tests we should create an additional test suite specifically for the jenkins recipe. You can take a peek at #7 which does this for the repo recipe.

metadata.rb
```diff
@@ -9,3 +9,5 @@
 
 issues_url 'https://github.com/nuclearsandwich/cookbook-ros-buildfarm/issues'
 source_url 'https://github.com/nuclearsandwich/cookbook-ros-buildfarm/'
+
+depends 'jenkins'
```
nuclearsandwich, line 13: You should add a version constraint here. Probably best to stick with the current major version (so use a constraint that limits us to any 8.*) https://docs.chef.io/config_rb_metadata/#version-constraints

attributes/jenkins.rb
```diff
@@ -0,0 +1,2 @@
+default['ros_buildfarm']['jenkins']['admin']['key'] = ''
```
nuclearsandwich: I think this will likely end up being a data bag item rather than an attribute since it is sensitive data.

recipes/jenkins.rb
```diff
@@ -0,0 +1,62 @@
...
+
+node.default['ros_buildfarm']['agent']['nodename'] = "agent_on_master"
+node.default['ros_buildfarm']['agent']['executors'] = 1
+node.default['ros_buildfarm']['agent']['labels'] = ["agent_on_master"]
+
+include_recipe '::agent' 
+
+directory '/var/lib/jenkins/casc_configs'
+template '/var/lib/jenkins/casc_configs/jenkins.yaml' do
+  group node['jenkins']['master']['group']
```
nuclearsandwich, line 62: Is there a reason you didn't update the file owner as well?

nuclearsandwich (dismissed):

nuclearsandwich:
@Maanasa-Kotha can you please update the PR description so it more closely describes the current state of the pull request. It would be a good idea to include a summary of what the recipe configures.

PR #8, opened 2020-10-26
Add reprepro-updater / reprepro for hosting apt repositories.

nuclearsandwich:
This change includes everything to get reprepro-updater and the apt repositories deployed for a build farm.

This is, by intention, pretty much a straight update of the [ros::repo module](https://github.com/ros-infrastructure/buildfarm_deployment/blob/master/modules/profile/manifests/ros/repo.pp) from buildfarm_deployment. There are some intentional changes:

* nginx is used for http rather than Apache. This is not made because Apache httpd was insufficient but because nginx is what the current maintainer (myself) is familiar with operationally.

* Repository suite and architecture information has been re-organized.
In the buildfarm_deployment_configuration, yaml-based hiera configs made embedding and maintaining file contents directly in hiera a viable option and the one that was used. However, it's not clear to those who aren't comfortable with the ROS build farm which items in the config file can be changed by end users and which are intended to be "static". Both the reprepro-updater ini config file and the default import_upstream configuration file have been moved to templates and the specific suites, architectures, and component are now chef attributes (which are roughly analogous to scalar puppet module parameters overridden via hiera). Default attributes are set in the cookbook but they can be superceded by attributes in the deployment environment.

I was partway through the implementation of pulp when I came up against deadlines that caused me to pause that work. There is a commit on this PR which has been applied and reverted and you'll notice the redis configuration file which has no purpose in the current branch. My intent is to resume work on getting the pulp work incorporated into chef after completing the initial migration. Unless something about the current state of the pulp work appears to interfere with the other running components I think  it can be ignored during this review as there will be a chance to review it in the subsequent revisions.


files/upload_repo.bash
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
```
j-rivero, line 2:
```suggestion
set -e
```
stop the script on presence of errors

recipes/repo.rb
```diff
@@ -1,49 +1,186 @@
...
+    group agent_username
+    mode '0600'
+  end
+end
+
+# Configure GPG for reprepro
+# .gnupg/gpg.conf
+directory "/home/#{agent_username}/.gnupg" do
+  owner agent_username
+  group agent_username
```
j-rivero, line 75:
```suggestion
  group agent_username
  mode '0700'
```

recipes/repo.rb
```diff
@@ -1,49 +1,186 @@
...
+end
+cookbook_file "/home/#{agent_username}/.gnupg/gpg.conf" do
+  source 'gpg.conf'
+  owner agent_username
+  group agent_username
```
j-rivero, line 81:
```suggestion
  group agent_username
  mode '0600'
```

recipes/repo.rb
```diff
@@ -1,49 +1,186 @@
...
+end
+
+gpg_key = data_bag_item('ros_buildfarm_repository_signing_keys', node.chef_environment)
+directory "/home/#{agent_username}/.ssh" do
+  owner agent_username
+  group agent_username
```
j-rivero, line 88:
```suggestion
  group agent_username
  mode '0700'
```

recipes/repo.rb
```diff
@@ -1,49 +1,186 @@
...
+template "/home/#{agent_username}/.buildfarm/reprepro-updater.ini" do
+  source 'reprepro-updater.ini.erb'
+  owner agent_username
+  group agent_username
+  mode '0600'
+  variables Hash[
+    architectures: node['ros_buildfarm']['apt_repos']['architectures'],
+    signing_key: gpg_key['fingerprint'],
+    suites: node['ros_buildfarm']['apt_repos']['suites'],
+  ]
```
j-rivero: there is an `upstream_config` variable being used in the template that it is not explicit here. Might be a good idea for completeness to declare it and leave it blank if it is optional.

recipes/repo.rb
```diff
@@ -1,49 +1,186 @@
...
+end
+
+git "/home/#{agent_username}/reprepro-updater" do
```
j-rivero, line 134: I assume that although the repository does not have too much activity, we want chef to update to latest changes on every provisioning run.

recipes/repo.rb
```diff
@@ -1,49 +1,186 @@
...
+  ]
+end
+
+# Initialise repositories with reprepro
+%w(building testing main).each do |repo|
+  execute "/usr/bin/python3 /home/#{agent_username}/reprepro-updater/scripts/setup_repo.py ubuntu_#{repo} -c" do
+    environment 'PATH' => '/bin:/usr/bin', 'PYTHONPATH' => "/home/#{agent_username}/reprepro-updater/src", 'HOME' => "/home/#{agent_username}"
+    user agent_username
+    group agent_username
+    not_if "/usr/bin/python3 /home/#{agent_username}/reprepro-updater/scripts/setup_repo.py ubunut_#{repo} -q"
```
j-rivero:
```suggestion
    not_if "/usr/bin/python3 /home/#{agent_username}/reprepro-updater/scripts/setup_repo.py ubuntu_#{repo} -q"
```

files/upload_repo.bash
nuclearsandwich: I'd actually prefer not to make this change now. The script has specific exit code handling for the ssh command from upstream. I can move it into a separate PR though.

recipes/repo.rb
nuclearsandwich: Do you think it's important that the GPG *config* is not readable by other users?

nuclearsandwich: Same as the .gpg directory I think we'd want this to be `0700` no?

recipes/repo.rb
nuclearsandwich: Do you think it's important that the gpg config be unreadable by other system users?

recipes/repo.rb
nuclearsandwich: Same as the .gpg directory I think we'd want this to be `0700` no?

recipes/repo.rb
nuclearsandwich: The default behavior of this resource is to update.
From https://docs.chef.io/resources/git/

> :sync
>     Default. Update the source to the specified version, or get a new clone or checkout. This >  action causes a hard reset of the index and working tree, discarding any uncommitted changes.

recipes/repo.rb
nuclearsandwich: Well spotted.

recipes/repo.rb
nuclearsandwich: Thanks fixed in 3ade7d4

files/upload_repo.bash
j-rivero: The `-e` should prevent us for possible failures in the script and return a success exit code. We can move to a different PR, it can affect heavily the behavior if some minor problems exist.

recipes/repo.rb
j-rivero: With the security kiddie hat on, I think the opposite: do we need to open it more than 0600? A malicious attack with a new/other user could potentially manipulate files in that directory.

recipes/repo.rb
j-rivero: yes, sorry, I failed miserably in making my sentence a question: do we want to update on every chef run, don't we? (no changes required in that case, yes).

recipes/repo.rb
j-rivero: `0700` is executable bit no? In directories is being used to access items inside even if you can not list items inside, I think. Do we want that instead of complete blocking access to non user?

j-rivero (approved):
Minor discussion about permissions, does not seem important enough to block the PR. Looks good.

recipes/repo.rb
nuclearsandwich: Yeah the intent is to always pull the latest. Good to check. In some cases I will specify the  the default action when I want to be explicit about what I want. Do you think this is a scenario where that makes sense?

j-rivero: totally.

files/upload_repo.bash
nuclearsandwich: I fully agree that errexit / -e is better for any bash script. My concern is that if we *rely* on the script working during a failure right now and don't know it I don't want to change that during the migration window. Hence moving this out to anther PR.

recipes/repo.rb
nuclearsandwich: Executable bit on directories is required for the directory to be traversable. That is, in order to access foo/bar/baz both foo and bar need to be traversable.

recipes/repo.rb
nuclearsandwich: efd38a5

files/upload_repo.bash
nuclearsandwich: Branch is up to be rebased for a future PR. https://github.com/ros-infrastructure/cookbook-ros-buildfarm/tree/upload-script-errexit

PR #9, opened 2020-10-27
Add an smtp server for jenkins

nuclearsandwich:
This adds postfix for SMTP based on what we've been using for the ROS buildfarm over the past several years. I can't say I recommend this to anyone else. On our wishlist is to move to a mail service provider rather than administer an SMTP server solely for the buid farm operations.

Postfix is stable and reliable and my go to server for my personal mail domains but it's rather a lot of software to be running just to have a functional `sendmail` bin so I was originally planning to evaluate [opensmtpd](https://www.opensmtpd.org/) if we continue to need mail delivery on the host. I'd suggest anyone else who wants it to look there instead.

### DIFF
--- a/attributes/jenkins.rb
+++ b/attributes/jenkins.rb
@@ -2,4 +2,4 @@ default['ros_buildfarm']['jenkins']['timezone'] = 'America/Los_Angeles'
 default['ros_buildfarm']['jenkins']['auth_strategy'] = 'default'
 default['ros_buildfarm']['server_name'] = 'ros_buildfarm'
 default['ros_buildfarm']['admin_email'] = 'noreply@ros_buildfarm'
-
+default['ros_buildfarm']['smtp'] = false
--- a/recipes/_jenkins_smtp.rb
+++ b/recipes/_jenkins_smtp.rb
@@ -1,0 +1,56 @@
+package 'postfix'
+package 'opendkim'
+package 'opendkim-tools'
+
+mail_name = node['ros_buildfarm']['server_name']
+file '/etc/mailname' do
+  content mail_name
+end
+template '/etc/postfix/main.cf' do
+  source 'postfix/main.cf.erb'
+  variables Hash[
+    hostname: mail_name
+  ]
+end
+
+dkim_config = data_bag_item('ros_buildfarm_dkim_configuration', node.chef_environment)
+
+directory "/etc/opendkim/keys/#{mail_name}" do
+  recursive true
+end
+dkim_config['keys'].each do |name, data|
+  file "/etc/opendkim/keys/#{mail_name}/#{name}.private" do
+    content data['private']
+    mode '0600'
+  end
+  file "/etc/opendkim/keys/#{mail_name}/#{name}.txt" do
+    content data['txt']
+    mode '0600'
+  end
+end
+
+trusted_hosts = dkim_config['trusted_hosts']
+unless trusted_hosts.include? mail_name
+  trusted_hosts << mail_name
+end
+file '/etc/opendkim/TrustedHosts' do
+  content trusted_hosts.join("\n")
+  mode '0644'
+end
+
+template '/etc/opendkim/KeyTable' do
+  source 'opendkim/keytable.erb'
+  variables Hash[
+    keys: dkim_config['keys'],
+    mail_name: mail_name,
+  ]
+  mode '0644'
+end
+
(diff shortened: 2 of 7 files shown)


templates/postfix/main.cf.erb
```diff
@@ -0,0 +1,36 @@
+<%# vim:ft=pfmain -%>
+smtpd_banner = $myhostname ESMTP $mail_name (Ubuntu)
+biff = no
+
+append_dot_mydomain = no
+readme_directory = no
+
+#TLS
+
+smtpd_tls_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
```
j-rivero, line 10: just double checking: are the snakeoil certs the one you want to use for mail?

attributes/jenkins.rb
```diff
@@ -2,4 +2,4 @@
 default['ros_buildfarm']['jenkins']['auth_strategy'] = 'default'
 default['ros_buildfarm']['server_name'] = 'ros_buildfarm'
 default['ros_buildfarm']['admin_email'] = 'noreply@ros_buildfarm'
-
+default['ros_buildfarm']['smtp'] = false
```
j-rivero, line 5: I only found this attribute used in `recipes/jenkins.rb` and defined here. Seems like with this there is code that is not going to be used. Is this ok?

recipes/_jenkins_smtp.rb
```diff
@@ -0,0 +1,56 @@
...
+  content mail_name
+end
+template '/etc/postfix/main.cf' do
+  source 'postfix/main.cf.erb'
+  variables Hash[
+    hostname: mail_name
+  ]
+end
+
+dkim_config = data_bag_item('ros_buildfarm_dkim_configuration', node.chef_environment)
```
j-rivero, line 16: I don't see a `data_bags` directory in the root of the repo, only under `test/integration`. Is this correct?

j-rivero (commented):
If I run `kitchen test` I found the following problem:
```
-----> Converging <agent-ubuntu-2004>...
       Preparing files for transfer
       Preparing dna.json
       Resolving cookbook dependencies with Berkshelf 7.1.0...
       Removing non-cookbook files before transfer
       Preparing data_bags
       Preparing environments
       Preparing solo.rb
root@'s password:
```
is something at my end?

Code looks nice, just some double-check questions.

templates/postfix/main.cf.erb
nuclearsandwich: Yes; sort of. This is the Ubuntu-default config and the actual port 25 SMTP server for this host is only trusted by the local host. We need postfix for its sendmail bin which is used by Jenkins but not to actually receive any mail.

recipes/_jenkins_smtp.rb
nuclearsandwich: Since this is a cookbook repository and not a standalone chef repository, it does not include cookbook dependencies, environments, or data bags. The data bags within the test/integration directory are used by the kitchen tests. When this cookbook is used for a deployment data bags will need to be added to the chef server or chef repository to be used for that deployment.

templates/postfix/main.cf.erb
nuclearsandwich: > is something at my end?

Yeah, if I had to guess your default kitchen provider is doing something odd. We can check it in chat.

attributes/jenkins.rb
nuclearsandwich: The vast majority of ros_buildfarm deployments should not use this direct smtp approach (at least not without thinking hard about the implications) so this is off by default intentionally.
This attribute is flipped on in the kitchen tests for jenkins https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/9/files#diff-dadef6cb66a2d8c23a5ffa6d1fa70221742fb18c0826c887fbb15ece6ed4f5b6R27-R29 https://github.com/ros-infrastructure/cookbook-ros-buildfarm/commit/f640deadd8f3596bd38b7547267cab8042fef536 so this recipe gets tested.

j-rivero: nice. How is going to be enabled when running real production deployment? EDIT: answered myself, since this is a generic cookbook public repository, the ROS2 buildfarm attributes/enviroments are not here.

j-rivero (approved):

PR #10, opened 2020-10-30
Fixes for acme.sh handling in jenkins.

nuclearsandwich:
This includes two fixes.

* Without a --webroot arg acme.sh fails since it requires a server method.

* Issuing a certificate without --force-renew causes an error and we are actually fine skipping this step if the domain is already configured. This is somewhat brittle and may need to be made more robust if our acme.sh config needs to change. But until we have a need it's hard to say how we should structure the chef logic.


cottsay (approved):
I don't have much knowledge in this scope, but the description makes sense to me.

PR #16, opened 2020-11-02
Update plugin versions to accomodate security fixes.

nuclearsandwich:
Plugin updates based on security warnings feedback on our test farm.

As Jose noted dealing with things like this is one hurdle we'll have to cross in order to do
unattended chef runs on Jenkins.

## Notable changes 

* The update of the email extension (`email-ext`) triggered a change to the jenkins.yaml
config that was incompatible and caused a restart failure.

* The Copy Artifact (`copyartifact`) plugin has added additional security protections and the new version defaults to a [migration mode](https://github.com/jenkinsci/copyartifact-plugin/blob/master/README.adoc#CopyArtifactPlugin-MigrationMode) during which the protections are not in place. This PR will leave the migration mode on so that further work can assess and implement the necessary changes.


attributes/plugins.rb
```diff
@@ -6,21 +6,21 @@
...
-  "cloudbees-folder" => "6.6",
+  "cloudbees-folder" => "6.14",
   "collapsing-console-sections" => "1.8.0",
-  "configuration-as-code" => "1.43",
   "command-launcher" => "1.2",
   "conditional-buildstep" => "1.3.6",
   "config-file-provider" => "3.6.2",
-  "copyartifact" => "1.41",
+  "configuration-as-code" => "1.43",
+  "copyartifact" => "1.45.2",
```
j-rivero, line 23: There is a disruptive change in 1.44 of copy artifact, [see changelog](https://github.com/jenkinsci/copyartifact-plugin/blob/master/CHANGELOG.adoc) although a migration mode is enabled by default and things should keep working. Not sure if you need to do any action.

nuclearsandwich: Yeah I've already looked at the migration mode. It would have been smart of me to mention as much in the PR description :+1:. I'll add info there. We're good in migration mode for now.

nuclearsandwich: Updated the PR description.

j-rivero (approved):

PR #19, opened 2020-11-05
Add gpg-vault service.

nuclearsandwich:
This is the adaptation of @cottsay's work creating an isolated gpg-vault
user running a shareable gpg-agent session.

Implementation note:
For other systemd services we've been installing the unit file and
triggering systemctl-daemon-reload manually. I noticed that chef has a
systemd_unit resource [1] and decided to try and use that to bundle
triggering the reload along with starting and enabling the service.

I'm interested in whether others like it better or not. It bakes systemd
into our chef but so does manually executing systemctl commands.
Feedback welcome.

[1]: https://docs.chef.io/resources/systemd_unit

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -68,6 +68,67 @@ data_bag('ros_buildfarm_upload_keys').each do |id|
   end
 end
 
+# Configure gpg-vault
+user 'gpg-vault' do
+  manage_home true
+  comment 'GPG vault user'
+end
+execute 'gpg-init' do
+  command 'gpg -K'
+  environment 'HOME' => '/home/gpg-vault'
+  not_if { File.directory? '/home/gpg-vault/.gnupg' }
+  user 'gpg-vault'
+  group 'gpg-vault'
+end
+cookbook_file '/home/gpg-vault/.gnupg/gpg-agent.conf' do
+  source 'repo/gpg-agent.conf'
+  owner 'gpg-vault'
+  group 'gpg-vault'
+  mode '0600'
+end
+directory '/var/run/gpg-vault' do
+  owner 'gpg-vault'
+  group 'gpg-vault'
+  mode '0750'
+end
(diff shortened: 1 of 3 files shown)


recipes/repo.rb
```diff
@@ -68,6 +68,67 @@
...
+  environment 'HOME' => '/home/gpg-vault'
+  user 'gpg-vault'
+  group 'gpg-vault'
+  not_if "gpg --list-secret-keys #{gpg_key['fingerprint']}"
+end
+group 'gpg-vault' do
+  append true
+  members [agent_username]
+  action [:manage]
+end
```
cottsay, line 130: Does this need to be further up in the file, before [the group name] is referenced?

cottsay (approved):
I tested this, and it works as expected. Thanks!

recipes/repo.rb
nuclearsandwich: I think the question is can we reference the gpg-vault group in the above resources before we use this resource?

If I'm wrong then you can ignore this and I would need clarification on the question.

We can reference it above because the gpg-agent group is created when the gpg-vault user is created based on default login.defs per this extraction from the useradd(8) manual

```
-g, --gid GROUP
 The group name or number of the user's initial login group. The group name must exist.
 A group number must refer to an already existing group.

 If not specified, the behavior of useradd will depend on the USERGROUPS_ENAB variable in
 /etc/login.defs.
 If this variable is set to yes (or -U/--user-group is specified on the command line), a
 group will be created for the user, with the same name as her loginname. If the variable is
 set to no (or -N/--no-user-group is specified on the command line), useradd will set the
 primary group of the new user to the value specified by the GROUP variable in /etc/default/useradd,
 or 100 by default.
```

cottsay: Yes, you answered my question. I understand the purpose of this snip now, thanks.

PR #22, opened 2020-12-14
Implement CI testing for kitchen test on Github Actions

j-rivero:
Refactor of #18 to bring only testing support into the PR

PR #24, opened 2020-12-14
Use upstream Jenkins cookbook.

nuclearsandwich:
We were using a fork in order to get a fix that I made previously but
that fix is now integrated into upstream in version 8.0.4.


nuclearsandwich:
I'm going to solo-merge this so I can rebase and merge #22 in order to get CI for some upcoming PRs.

PR #25, opened 2020-12-14
Enable per-environment ros_buildfarm_jenkins_scripts for auth.

nuclearsandwich:
Different environments may require different auth logic.
With this change it is now required for the auth_script data bag item to have scripts nested under a key named for the chef environment.


j-rivero:
>  With this change it is now required for the auth_script data bag item to have scripts nested under a key named for the chef environment.

Is this a change that needs to happen somehow in this repository?

nuclearsandwich:
> Is this a change that needs to happen somehow in this repository?

Nope, the cookbook defaults don't use the groovy script method to configure authentication. The change supporting this in our chef configuration repository has already been made.

j-rivero (approved):

PR #28, opened 2020-12-15
Add container registry cache on repository host.

nuclearsandwich:
This PR adds a configured a pull-through Docker registry cache on the repository host.
The cache is configured by default but is not used by any ros_buildfarm nodes unless a corresponding address is added to the `node['docker']['registry_mirrors']` attribute.

There is another PR coming which adds a customized seccomp profile that will almost certainly conflict with this one. It doesn't really matter which one lands first but let's go in order of first-opened-first-merged and I'll rebase that PR once this one has merged.

### DIFF
--- a/recipes/agent.rb
+++ b/recipes/agent.rb
@@ -11,6 +11,13 @@ file '/etc/containerd/config.toml' do
   content 'disabled_plugins = ["cri"]'
   notifies :restart, 'service[containerd]'
 end
+template '/etc/docker/daemon.json' do
+  source 'docker-daemon.json.erb'
+  variables Hash[
+    registry_mirrors: node['docker']['registry_mirrors']
+  ]
+  notifies :restart, 'service[docker]'
+end
 
 agent_username = node['ros_buildfarm']['agent']['agent_username']
 agent_homedir = "/home/#{agent_username}"
--- a/recipes/agent.rb
+++ b/recipes/agent.rb
@@ -1,11 +1,5 @@
 apt_update
 package 'docker.io'
-service 'docker' do
-  action [:start, :enable]
-end
-service 'containerd' do
-  action :nothing
-end
 directory '/etc/containerd'
 file '/etc/containerd/config.toml' do
   content 'disabled_plugins = ["cri"]'
@@ -17,6 +11,12 @@ template '/etc/docker/daemon.json' do
     registry_mirrors: node['docker']['registry_mirrors']
   ]
   notifies :restart, 'service[docker]'
+end
+service 'containerd' do
+  action :nothing
+end
+service 'docker' do
+  action [:start, :enable]
 end
 
 agent_username = node['ros_buildfarm']['agent']['agent_username']
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -272,3 +272,15 @@ if not node['ros_buildfarm']['repo']['rsyncd_endpoints'].empty?
     action [:start, :enable]
   end
 end
+
+# Configure read-through container registry cache
+if node['ros_buildfarm']['repo']['container_registry_cache_enabled']
+  package 'docker-registry'
+  cookbook_file '/etc/docker/registry/config.yml' do
+    source 'docker-registry-config.yml'
+    notifies :restart, 'service[docker-registry]'
+  end
+  service 'docker-registry' do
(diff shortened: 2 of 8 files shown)


recipes/agent.rb
```diff
@@ -1,5 +1,15 @@
 apt_update
 package 'docker.io'
+service 'docker' do
```
j-rivero: can the service be added after the template to save the `notifies` or is there any benefit in starting it first and restart just after deploying the configuration?

recipes/repo.rb
```diff
@@ -272,3 +272,15 @@
     action [:start, :enable]
   end
 end
+
+# Configure read-through container registry cache
+if node['ros_buildfarm']['repo']['container_registry_cache_enabled']
+  package 'docker-registry'
+  cookbook_file '/etc/docker/registry/config.yml' do
+    source 'docker-registry-config.yml'
+    notifies :restart, 'service[docker-registry]'
```
j-rivero, line 281: Do this restart just before the service do anything different than just remove it?

templates/docker-daemon.json.erb
```diff
@@ -0,0 +1,4 @@
+{
+  <%= if not @registry_mirrors.empty? then %("registry-mirrors": [ ) + @registry_mirrors.map{|m| %("#{m}")}.join(", ") + " ],\n" end -%>
+  "seccomp-profile": "/etc/docker/seccomp/default-with-personality.json"
```
j-rivero: This file is going to be added in an upcoming PR, right?

j-rivero (commented):
Looks good, minor comments

recipes/agent.rb
nuclearsandwich: In general yes, I try to order recipes with package, config files, service.

Due to the development order I didn't do that here and should. But we can't remove the notification because we want to make sure that a change to the config file triggers a restart if the docker service is already running.

recipes/repo.rb
nuclearsandwich: notification only occurs when the file is changed. So when chef is run and this file doesn't change, it won't restart the service. But if the file does change and the service is always running, without the notification to restart the docker-registry service wouldn't pick up the new config.

templates/docker-daemon.json.erb
nuclearsandwich: Right, I didn't mean to commit this.

nuclearsandwich: It's been rebased out.

recipes/agent.rb
nuclearsandwich: Services have been moved below configuration in the latest patch.

recipes/repo.rb
j-rivero: thanks for the explanation.

j-rivero:
Looks good to me. It would be nice to add a docker operation (might be a single pull) in the tests so we can check that the cache is not heavily broken.

nuclearsandwich:
> It would be nice to add a docker operation (might be a single pull) in the tests so we can check that the cache is not heavily broken.

As has been discussed before, I do not see the appeal of a lot of the sample infrastructure tests which just express the same information as the configuration but as constraints rather than declarations. However, at your prompting I added tests that verify the output of `docker info` and `docker run hello-world` and in fact caught a bug introduced by splitting the seccomp profile changes out of this PR. However, we can't run docker inside the dokken runners on GitHub Actions so I'll have to figure out how to mask out those tests.

nuclearsandwich:
>  I'll have to figure out how to mask out those tests.

I couldn't figure out a way to detect running within kitchen-dokken in short order so I've commented those tests out entirely for now.

j-rivero:
> I couldn't figure out a way to detect running within kitchen-dokken in short order so I've commented those tests out entirely for now.

Disabling it if we detect that code is running on a docker container (checking for `/.dockerenv` file) will disable it in more occasions that it should but will improve the situation where the code is always disabled.

j-rivero (approved):

nuclearsandwich:
> Disabling it if we detect that code is running on a docker container (checking for `/.dockerenv` file) will disable it in more occasions that it should but will improve the situation where the code is always disabled.

That's a good idea. I'll keep it in mind as a way to bring those tests back in the future.

PR #31, opened 2020-12-15
Nest Jenkins attributes under a ['jenkins'] object and add warnings for old attribute hierarchy.

nuclearsandwich:
This was done inconsistently in some resources and not others causing improper configuration.
Both of these attributes are specific to the jenkins recipe and so are nested under a jenkins object.

I've also attempted to pilot a warning system for how I'd like to propagate changes to expected attributes in the future.

### DIFF
--- a/attributes/jenkins.rb
+++ b/attributes/jenkins.rb
@@ -1,5 +1,5 @@
 default['ros_buildfarm']['jenkins']['timezone'] = 'America/Los_Angeles'
 default['ros_buildfarm']['jenkins']['auth_strategy'] = 'default'
-default['ros_buildfarm']['server_name'] = 'ros_buildfarm'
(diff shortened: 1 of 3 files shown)


j-rivero (commented):
Do we need to update the [smtp recipe](https://github.com/ros-infrastructure/cookbook-ros-buildfarm/blob/latest/recipes/_jenkins_smtp.rb#L5) too?

nuclearsandwich:
> Do we need to update the [smtp recipe](https://github.com/ros-infrastructure/cookbook-ros-buildfarm/blob/latest/recipes/_jenkins_smtp.rb#L5) too?

Well spotted!

j-rivero (approved):
as soon as the CI finished green, ready to go

PR #33, opened 2020-12-15
Purge fingerprints directory on reconfiguration.

nuclearsandwich:
Credential files are recreated on each chef run in a non-idempontent
fashion. This may just be a quirk of the cookbook or it could be a
Jenkins limitation due to the per-instance encryption of sensitive data.
Either way we live with it for now by unconditionally purging the
fingerprints directory.

### DIFF
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -249,6 +249,31 @@ data_bag('ros_buildfarm_password_credentials').each do |item|
   end
 end
 
+# Remove Jenkins fingerprint files
+# Jenkins tracks the fingerprints of certain files so that different jobs may use different versions.
+# The chef credential resources do not seem idempotent in the face of existing credentials so when chef
+# is run and they are updated the fingerprint of the credential changes which causes an issue with jobs
+# using them.
+# I would like to find a more elegant solution, either of getting Jenkins to use the current version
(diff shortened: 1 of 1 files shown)


j-rivero (commented):
I'm not sure about the implications of this change. The fingerprint directory contains hashes of files used by builds. If we remove it once applying an update to a buildfarm, are we destroying critical information about the builds that have run on it?

nuclearsandwich:
> it once applying an update to a buildfarm, are we destroying critical information about the builds that have run on it?

For some values of "critical", yes. But for the ROS buildfarm I'm not aware of any situation actually using these fingerprints aside from the credential files which are breaking as a result.

j-rivero:
> For some values of "critical", yes. But for the ROS buildfarm I'm not aware of any situation actually using these fingerprints aside from the credential files which are breaking as a result.

Reading a[ bit about the fingerprints](https://thingsyoudidntknowaboutjenkins.tumblr.com/post/24737092433/fingerprinting) seems to me like it could be build related information (particularly about files belonging to every build) that could be broken if we remove the fingerprints database. Might not be critical to work but could affect part of the old builds. Let's be sure of testing this well before deploying.

j-rivero (approved):

PR #35, opened 2020-12-17
Pin to version 8.0.4.

nuclearsandwich:
The change introduced in 8.1.0 breaks the use of run_state supplied
usernames and passwords for CLI. I need to ticket the issue upstream but
I want to find out a bit more about why the change was introduced first.


nuclearsandwich:
I'm going to YOLOmerge this pin back so I can rebase the PRs with CI blocked by this issue.

PR #36, opened 2020-12-19
Import the new signing key for the ROS bootstrap repository and check it by default

nuclearsandwich:
This adds the public signing key for the ROS bootstrap repository on repos.ros.org to this cookbook. It also enables checking that key by default. Disabling to stick with the previous behavior can be done by setting the signing key attribute to `blindtrust` instead of a key id.


ros-discourse:

https://discourse.ros.org/t/ros-bootstrap-repository-is-now-signed/18106/1

recipes/repo.rb
```diff
@@ -181,15 +181,33 @@
...
 execute "gpg --import /home/#{agent_username}/.ssh/gpg_private_key.sec" do
   user agent_username
   group agent_username
   environment 'PATH' => '/bin:/usr/bin', 'HOME' => "/home/#{agent_username}"
-  not_if "gpg --list-secret-keys | grep #{gpg_key['fingerprint']}"
+  not_if "gpg --list-secret-keys #{gpg_key['fingerprint']}"
+end
+
+# Import ROS bootstrap signing key for signature verification
+cookbook_file "/home/#{agent_username}/.ssh/ros_bootstrap.pub.asc" do
```
j-rivero, line 194: Does this work without a source property that points to `ros_boostrap.pub.asc`?

j-rivero (commented):
One doubt: the cookbook is public and designed to be reused. What is the way for an user of this cookbook to define his/her own keys? Needs to overwrite the hardcoded values?

recipes/repo.rb
nuclearsandwich: Chef has default search behavior for template and file sources. I'm not 100% convinced we should rely on the implicit defaults so I would change this if you prefer explicit to implicit.

https://docs.chef.io/resources/cookbook_file/

nuclearsandwich:
> What is the way for an user of this cookbook to define his/her own keys? Needs to overwrite the hardcoded values?

You're correct that right now using a custom bootstrap repository is missing a step: trusting the public key. I opened an enhancement issue #38 which details a workaround and why I have waited on implementing it fully.

recipes/repo.rb
j-rivero: did not know about the default behaviour before this PR, I assume it since this was working. It is fine, don't see any problem or need to change it right now.

j-rivero (approved):

PR #39, opened 2021-01-14
Add a custom seccomp profile for Docker.

nuclearsandwich:
This is a version of Docker's default seccomp profile from moby/moby v20.10.2 with a modification to add the personality system call used by sbcl in ROS 1.


ros-discourse:

https://discourse.ros.org/t/ssl-failure-in-buildfarm-jobs-ros1/18304/6

cottsay (approved):
I know next to nothing about seccomp, but the spirit of this change makes sense.

PR #40, opened 2021-01-20
Add Docker cleanup script.

nuclearsandwich:
This script is less smart than the old docker cleanup script but is also
a four line shell script. It will run on the hour and only clear docker
images when necessary.
As a tradeoff for simplicity it will clear everything that is not in
active use.


cottsay (approved):
You might consider dropping some of the three consecutive empty lines being added to `agent.rb`, but LGTM other than that.

nuclearsandwich:
> You might consider dropping some of the three consecutive empty lines being added to `agent.rb`, but LGTM other than that.

Done in an amended commit.

PR #41, opened 2021-01-21
Add pulp client packages to agent.

nuclearsandwich:
These packages are used to connect to the pulp service.
There is no built in chef resource for python packages from pypi and the
options for python provider cookbooks seem quite stale in Chef
supermarket. With the use of exact versions the pip install command is
idempotent and pypi will uninstall any other versions of the package as
we update in the future.
As a result I think sticking with an execute resource rather than trying
to import a dependency is the way to go for now.

### DIFF
--- a/recipes/agent.rb
+++ b/recipes/agent.rb
@@ -85,6 +85,12 @@ end
   package vcspkg
 end
 
+# Client utilities for pulp content manager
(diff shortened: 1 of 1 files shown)


cottsay (approved):
Thank you!

PR #42, opened 2021-01-22
Use environment-specific private key credentials.

nuclearsandwich:
Private key credentials were not previously using environment-specific credentials but they should be.

This also adds an explicit name field since the credential name may not match the data bag ID.

### DIFF
--- a/test/integration/data_bags/ros_buildfarm_private_key_credentials/private.json
+++ b/test/integration/data_bags/ros_buildfarm_private_key_credentials/private.json
@@ -1,6 +1,7 @@
 {
   "id": "private",
   "test": {
+    "name": "jenkins-agent",
     "description": "a private key",
     "private_key": "-----BEGIN RSA PUBLIC KEY-----\nMIGJAoGBAMQEEu72Tn1D8mZfUiGppAlZsmSabFaZmoNC+k9T4PaeVRrtKIM5ZPCr\nTYyzrNngIXfVodlOxoRrzCUlt7fhLRjB6p6DqlfXr74BdzMnnq+/5YTYDWDfWo3M\nCpMKFa1XIcjZ6x4uFx/CQhx/qOmvnibPlOOFa6MHRDh1hVFuXZ7BAgMBAAE=\n-----END RSA PUBLIC KEY-----\n"
   }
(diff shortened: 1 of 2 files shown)


test/integration/data_bags/ros_buildfarm_private_key_credentials/private.json
```diff
@@ -1,5 +1,8 @@
 {
   "id": "private",
-  "description": "a private key",
-  "private_key": "-----BEGIN RSA PUBLIC KEY-----\nMIGJAoGBAMQEEu72Tn1D8mZfUiGppAlZsmSabFaZmoNC+k9T4PaeVRrtKIM5ZPCr\nTYyzrNngIXfVodlOxoRrzCUlt7fhLRjB6p6DqlfXr74BdzMnnq+/5YTYDWDfWo3M\nCpMKFa1XIcjZ6x4uFx/CQhx/qOmvnibPlOOFa6MHRDh1hVFuXZ7BAgMBAAE=\n-----END RSA PUBLIC KEY-----\n"
+  "test": {
+    "name": "jenkins-agent",
+    "description": "a private key",
+    "private_key": "-----BEGIN RSA PUBLIC KEY-----\nMIGJAoGBAMQEEu72Tn1D8mZfUiGppAlZsmSabFaZmoNC+k9T4PaeVRrtKIM5ZPCr\nTYyzrNngIXfVodlOxoRrzCUlt7fhLRjB6p6DqlfXr74BdzMnnq+/5YTYDWDfWo3M\nCpMKFa1XIcjZ6x4uFx/CQhx/qOmvnibPlOOFa6MHRDh1hVFuXZ7BAgMBAAE=\n-----END RSA PUBLIC KEY-----\n"
```
j-rivero, line 6: I assume that we are using this file as an stub for testing but seems weird to me to find a public key (is it a fake?) where there should be a private key.

nuclearsandwich: All of the keys or secrets in the test/integration/data_bags file are throwaway. Some tests can have any old junk in the data bags but some are processed such that the test data must be a key of valid form.

j-rivero (approved):

PR #43, opened 2021-01-22
Set HOME environment variable in case chef user is not root.

nuclearsandwich:
This resolves some strange configuration problems caused by running chef via sudo as another user instead of in a root environment.

### DIFF
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -225,6 +225,7 @@ if node['ros_buildfarm']['letsencrypt_enabled']
     not_if 'test -d /root/acme.sh'
   end
   execute 'acmesh-install' do
+    environment 'HOME' => '/root'
     cwd '/root/acme.sh'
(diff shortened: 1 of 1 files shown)


recipes/jenkins.rb
```diff
@@ -225,6 +225,7 @@ def jenkins = Jenkins.getInstance()
     not_if 'test -d /root/acme.sh'
   end
   execute 'acmesh-install' do
+    environment 'HOME' => '/root'
```
j-rivero, line 228: Could be an alternative solution to use the `user` property and set it to `root`?

nuclearsandwich: On Linux, the user/group field will change the effective user/group ID but does not set up the environment as if that user is logged in and running the command.

I tend not to specify `user "root"` in recipes that would not work if they are run directly as a non-root user.

j-rivero (approved):

PR #44, opened 2021-01-28
Add documentation annotations to attributes.

nuclearsandwich:
Starting off the documentation efforts with the explanation of each attribute, its uses, and any restrictions on its values. The plan is to link the attributes directory directly from more walkthrough-style documentation in the chef-ros-buildfarm repository.


attributes/agent.rb
```diff
@@ -1,7 +1,31 @@
+# Linux username which will run Jenkins Java agent process.
+# This user is entirely managed by the cookbook.
 default['ros_buildfarm']['agent']['agent_username'] = 'jenkins-agent'
```
cottsay, line 3: From a security perspective, I think we should mention that this user has signing capabilities and is also accessible by the worker hosts.

attributes/default.rb
```diff
@@ -1,2 +1,7 @@
+# The public URL for your Jenkins instance
+# This may be an IP address or domain name but must be resolvable and reachable from all agents.
 default['ros_buildfarm']['jenkins_url'] = ''
+
+# Docker is installed on every machine for use in Jenkins.
+# If your configuration includes a docker registry cache or mirror you can configure that here.
 default['docker']['registry_mirrors'] = []
```
cottsay, line 7: What is expected here? Some sort of URL, or a hostname? Maybe including a port?

**EDIT:** I see an example in the docs for `default['ros_buildfarm']['repo']['container_registry_cache_enabled']`. Might be nice to have a reference or example here anyway.

cottsay (approved):
A couple of suggestions, but this looks great.

PR #45, opened 2021-02-01
Store Jenkins token in a file rather than environment

nuclearsandwich:
Reduce the ease with which a Jenkins token may be captured by a
malicious Jenkins job.
Additional measures should be taken:
1. Use a Jenkins user which only has the necessary permissions for agent
creation and configuration and not full admin access.
2. Restrict Jenkins agent connections to trusted networks.
3. Switch to a Jenkins API token rather than password to prevent a
captured token from being used to authenticate via the Jenkins web UI.

### DIFF
--- a/kitchen.yml
+++ b/kitchen.yml
@@ -1,4 +1,6 @@
 ---
+driver:
+  vagrant
 provisioner:
   name: chef_solo
   product_name: chef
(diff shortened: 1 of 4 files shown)


j-rivero (approved):
Looks perfect to me. In fact I opened a [PR in osrf_jenkins_agent](https://github.com/osrf/osrf_jenkins_agent/pull/12) to fix the problems with `/etc/default/jenkins-agent` with similar permissions. In that PR I assumed that the init system will read  `/etc/default/jenkins-agent` as root (was probably incorrect).

PR #48, opened 2021-02-06
Run pulp content repository on the repo host.

nuclearsandwich:
Run the [pulp](https://pulpproject.org) content repository software in containers on the repo host.
This branch uses docker for the contained services. I struggled for a while to try and get a podman based approach in order to leverage rootless and daemonless containers but ran into constant challenges with the effective user and environment which where not yielding to reasonable seeming solutions and I was running out of motivation to figure out what the heck was going on. The abandoned pulp branch is https://github.com/ros-infrastructure/cookbook-ros-buildfarm/tree/nuclearsandwich/podman-pulp-wip

In an earlier iteration of this project the original repository for systemd-docker was not working and unmaintained. I found a maintained fork and have been successfully using the binary provided there.


nuclearsandwich:
@cottsay if the setup here looks reasonably complete I'll go ahead and deploy it to the test build farm. I'm aware there are still some TODOs about getting the gpg stuff set up for pulp and lingering comments (of mine) for creating pulp workers and creating the rpm repos. I think the workers comment is probably stale given that pulp workers are created just above it but I haven't returned to https://github.com/ros-infrastructure/buildfarm_deployment/pull/228/ to port over configuring and creating the RPM repositories.

nuclearsandwich:
The CI failures are due to the action running test kitchen inside docker. @j-rivero do you know whether dokken allows you to bind-mount the host docker socket so that sibling containers can be spawned?

files/rpm_repo_init.py
```diff
@@ -0,0 +1,220 @@
...
+        new_names = set()
+
+        for pkg in packages:
+            if [r for r in pkg.requires if r[0] in target_names]:
+                to_remove[pkg.pulp_href] = pkg
+                new_names.add(pkg.name)
+
+    return to_remove
+
+
```
cottsay:
```suggestion
```

Looks like this can actually be dropped. I don't see it referenced.

kitchen.dokken.yml
```diff
@@ -5,6 +5,7 @@ driver:
   use_sudo: false
   privileged: true
   env: [CHEF_LICENSE=accept]
+  volumes: [ '/var/lib/docker', '/var/run/docker.sock' ]
```
cottsay:
```suggestion
  volumes: ['/var/lib/docker', '/var/run/docker.sock']
```

templates/pulp/pulp.service.erb
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=<%= @description %>
+After=docker.service <%= @after_units.join(' ') %>
+Require=docker.service <%= @required_units.join(' ') %>
+
+[Service]
+ExecStart=/usr/local/bin/systemd-docker --cgroups name=systemd run --rm <%= @docker_create_args %> --name %p%i  pulp_image <%= @docker_cmd %>
```
cottsay:
```suggestion
ExecStart=/usr/local/bin/systemd-docker --cgroups name=systemd run --rm <%= @docker_create_args %> --name %p%i pulp_image <%= @docker_cmd %>
```

recipes/repo.rb
```diff
@@ -265,6 +265,162 @@
...
+  mode '0700'
+end
+directory "#{pulp_data_directory}/media" do
+  owner 'pulp'
+end
+
+cookbook_file "#{pulp_data_directory}/initialize.py" do
+  owner 'pulp'
+  source 'rpm_repo_init.py'
+  mode '0600'
```
cottsay, line 291: If we're going to execute the script explicitly with the interpreter, we should drop the shebang from `rpm_repo-init.py`.

recipes/repo.rb
```diff
@@ -265,6 +265,162 @@
...
+  command 'docker run --user 1200 --rm -v /var/run/postgresql:/var/run/postgresql pulp_image pulpcore-manager migrate --noinput'
+end
+
+execute 'pulp_collect_static' do
+  command "docker run --user 1200 --rm -v #{pulp_data_directory}:/var/repos/.pulp pulp_image pulpcore-manager collectstatic --noinput"
+end
+
+pulp_admin_password = data_bag_item('ros_buildfarm_pulp_admin_password', node.chef_environment)['password']
+execute 'set_pulp_admin_password' do
+  command "docker run --user 1200 --rm -v /var/run/postgresql:/var/run/postgresql pulp_image pulpcore-manager reset-admin-password -p '#{pulp_admin_password}'"
```
cottsay: Note that this step was missing shell escaping for the password in the original PR. The randomly generated password I used on the test farm broken the step.

recipes/repo.rb
```diff
@@ -265,6 +265,162 @@
...
+  action :nothing
+end
+
+template "/etc/systemd/system/pulp-api-endpoint.service" do
+  source "pulp/pulp.service.erb"
+  variables Hash[
+    description: "Pulp API Endpoint",
+    after_units: %w(postgresql.service redis-server),
+    required_units: %w(postgresql.service redis-server),
+    docker_create_args: %(-u 1200 -v #{pulp_data_directory}:/var/repos/.pulp -v /var/run/postgresql:/var/run/postgresql -v /var/run/redis:/var/run/redis -p 24817:24817),
```
cottsay:
```suggestion
    docker_create_args: %(-u 1200:1200 -v #{pulp_data_directory}:/var/repos/.pulp -v /var/run/postgresql:/var/run/postgresql -v /var/run/redis:/var/run/redis -p 24817:24817),
```

cottsay (commented):
A few comments, but this looks like a good start, and the bulk of what is needed to move forward.

recipes/repo.rb
nuclearsandwich: Do you have a preference for executable vs otherwise? I don't.

recipes/repo.rb
nuclearsandwich: Right. I had dropped the : separator because in the podman work `1200:1200` indicates run the process with uid 1200 inside the container mapped to uid 1200 on the host whereas in docker it indicates the usual uid:gid.

recipes/repo.rb
cottsay: In the spirit of deterministic behavior, I'd guess my preference would be to invoke it using the interpreter directly, and not to rely on the shebang. But I don't really care very much either, just looking for consistency.

recipes/repo.rb
```diff
@@ -265,6 +265,163 @@
...
+
+  execute 'pulp_django_migration' do
+    command 'docker run --user 1200 --rm -v /var/run/postgresql:/var/run/postgresql pulp_image pulpcore-manager migrate --noinput'
+  end
+
+  execute 'pulp_collect_static' do
+    command "docker run --user 1200 --rm -v #{pulp_data_directory}:/var/repos/.pulp pulp_image pulpcore-manager collectstatic --noinput"
+  end
+
+  pulp_admin_password = data_bag_item('ros_buildfarm_pulp_admin_password', node.chef_environment)['password']
```
cottsay: It would be good to get this credential populating on the Jenkins master. Once we have that and the pulp client pip packages installed on the master and all of the workers, the remaining changes will all be on the repo host.

nuclearsandwich: Updated in 
2987f5a

cottsay (approved):
I think this is a good/stable milestone.

It wouldn't be a bad idea to merge this as-is, deploy it to the staging farm, and iterate from there.

recipes/agent.rb
```diff
@@ -100,7 +100,7 @@
 
 # Client utilities for pulp content manager
 package "python3-pip"
-%w[pulp-rpm-client==3.8.0 pulpcore-client==3.8.1].each do |pulppkg|
+%w[pulp-rpm-client==3.4.1 pulpcore-client==3.4.1].each do |pulppkg|
```
cottsay: I did some testing, and it looks like the issues we were having with the newer Pulp versions were caused by a mismatch, and not necessarily our use of the API. Specifically, things seem to go awry if the client code is newer than the API server.

I found that `pulpcore==3.9.1` and `pulp-rpm==3.9.0` (and their corresponding `-client` packages) seemed to work well in concert.

It would probably be worth upgrading everything prior to merging this and capturing the new AMI for the workers.

nuclearsandwich: By upgrading everything do you mean updating the pulp version both on the agents and in the pulp image?

cottsay: Correct

PR #49, opened 2021-02-26
Use xslt to customize autoindex pages.

nuclearsandwich:
The default nginx autoindex listing has hard length limits on the displayed files. This is a basic recreation of it without limits on the file length.


j-rivero (approved):
Changes looks to me. If you test them somehow online/offline, the PR is good to go IMHO.

nuclearsandwich:
> Changes looks to me. If you test them somehow online/offline, the PR is good to go IMHO.

I developed the stylesheet using the staging ROS buildfarm deployment but those manual changes have since been reverted, this will reintroduce them and perhaps iteration will be required.

PR #50, opened 2021-02-26
Add reverse proxy configuration for pulp RPM content.

nuclearsandwich:
This patch adds reverse-proxying pulp content endpoints for the RPM building repository. The plan is to eventually render the testing and main repositories on the filesystem but they could also be made to use this approach if needed.

While I had originally intended to do this with regexp directives and matching, since we have the repository configuration data we can actually pre-render location directives as a template. There's a copy-pasted TODO for adding more reverse proxy directives which I can actually do as a config block include. Also on the TODO list is adding caching to repository data files to prevent failures when metadata files change non-atomically.


cottsay (approved):
:tada:

PR #51, opened 2021-02-27
Prevent shenanigans with gpg-vault and pulp users

cottsay:
Make them 'system' accounts (uid < 1000) and take away their shell.


nuclearsandwich (approved):
I tested how this change would behave when updating a running system and while the UIDs won't change if the user already exists it will update the shell.

PR #54, opened 2021-02-27
Add new 'group_execute' resource

cottsay:
The Chef 'execute' resource allows you to specify user and group permissions with which to execute the command. Unfortunately, this doesn't take into account seccondary groups listed in /etc/group.

We're currently using those seccondary groups to enable a user (process) to access the GPG vault, and the reprepro commands invoked by chef will need to perform some signing, and will therefore need permission.

The new resource defined by this change is nearly identical to the 'execute' resource, but adds an additional 'seccondary_groups' property. The resource then augments the supplied command by prefixing it with calls to 'sg <group-name>' to chain through the given groups. If a 'group' value was specified by the resource, that group is added to the call chain last to ensure it is the primary group after the other groups have been added.

Some examples, assuming that `jenkins-agent` is a member of the `gpg-vault` group:

A normal `execute` invocation:
```
execute 'id' do
  user 'jenkins-agent'
  group 'jenkins-agent'
end
```
Result: `uid=1001(jenkins-agent) gid=1001(jenkins-agent) groups=1001(jenkins-agent),0(root)`

An invocation using `group_execute`:
```
group_execute 'id' do
  user 'jenkins-agent'
  group 'jenkins-agent'
  secondary_groups ['gpg-vault']
end
```
Result: `uid=1001(jenkins-agent) gid=1001(jenkins-agent) groups=1001(jenkins-agent),0(root),997(gpg-vault)`
The actual command invoked is: `sg gpg-vault sg\ jenkins-agent\ id`

### DIFF
--- a/libraries/group_execute_provider.rb
+++ b/libraries/group_execute_provider.rb
@@ -1,0 +1,28 @@
+class Chef
+  class Provider
+    class GroupExecute < Chef::Provider::Execute
+      provides :group_execute
+
+      def_delegators :secondary_groups
+
+      private
+
+      def command
+        command = new_resource.command
+
+        if new_resource.secondary_groups
+          groups = new_resource.secondary_groups
+          if group
+            # Always do the 'primary' group last in the call chain
+            groups = [group] + groups
+          end
+          groups.each do |secondary_group|
+            command = "sg #{secondary_group} #{Shellwords.shellescape(command)}"
+          end
+        end
+
+        command
+      end
+    end
+  end
+end
--- a/libraries/group_execute_provider.rb
+++ b/libraries/group_execute_provider.rb
@@ -14,7 +14,7 @@ class Chef
           groups = new_resource.secondary_groups
           if group
             # Always do the 'primary' group last in the call chain
-            groups = groups + [group]
+            groups << group
           end
           groups.reverse.each do |secondary_group|
             command = "sg #{secondary_group} #{Shellwords.shellescape(command)}"
--- a/libraries/group_execute_provider.rb
+++ b/libraries/group_execute_provider.rb
@@ -14,9 +14,9 @@ class Chef
           groups = new_resource.secondary_groups
           if group
             # Always do the 'primary' group last in the call chain
-            groups = [group] + groups
+            groups = groups + [group]
           end
-          groups.each do |secondary_group|
+          groups.reverse.each do |secondary_group|
             command = "sg #{secondary_group} #{Shellwords.shellescape(command)}"
           end
         end
--- a/libraries/group_execute_resource.rb
+++ b/libraries/group_execute_resource.rb
@@ -1,0 +1,11 @@
+class Chef
+  class Resource
+    class GroupExecute < Chef::Resource::Execute
+      provides :group_execute, target_mode: true
+
+      def secondary_groups(arg=nil)
+        set_or_return(:secondary_groups, arg, :kind_of => Array)
+      end
+    end
+  end
+end
--- a/libraries/group_execute_resource.rb
+++ b/libraries/group_execute_resource.rb
@@ -3,6 +3,10 @@ class Chef
     class GroupExecute < Chef::Resource::Execute
       provides :group_execute, target_mode: true
 
+      def command(arg=nil)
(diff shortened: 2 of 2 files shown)


libraries/group_execute_provider.rb
```diff
@@ -0,0 +1,28 @@
...
+      private
+
+      def command
+        command = new_resource.command
+
+        if new_resource.secondary_groups
+          groups = new_resource.secondary_groups
+          if group
+            # Always do the 'primary' group last in the call chain
+            groups = [group] + groups
```
nuclearsandwich: Unless I'm misreading, it seems like the comment and statement disagree.
The comment says the primary group should be last but then concatenates an array containing just the primary group to the head of the array of groups.

libraries/group_execute_provider.rb
```diff
@@ -0,0 +1,28 @@
...
+          end
+          groups.each do |secondary_group|
+            command = "sg #{secondary_group} #{Shellwords.shellescape(command)}"
```
nuclearsandwich, line 20: Does this work as expected if the original command is an Array rather than a String?

libraries/group_execute_provider.rb
cottsay: The call stack is created from the inside out, which makes the shell escaping easier. So the last group in the list is the first one to get evaluated as it works in towards the actual command. The order of the groups seems to have no bearing on the behavior, beyond the fact that the last (inner-most) call ends up setting the primary group.

libraries/group_execute_provider.rb
cottsay: Oh, I remember seeing that was a possible type for the argument and then blocking it from my mind. No, it's not handled. Would it be OK if I just asserted it as unsupported? I don't really want to deal with escaping spaces in arguments prior to joining the array into a flat string.

nuclearsandwich: Yeah I think as long as it blows up rather than silently does something unexpected that's fine.

In vanilla Ruby I'd just raise an ArgumentError but I'm not sure if there is a more chefly way to re-declare an inherited parameter. If there's nothing obvious in the docs you've already encountered I'd suggest just dumping the following at the start of the command method.
```ruby
raise ArgumentError, "Only String commands are supported" unless command.is_a? String
```

cottsay: I think I might know a way, let me give it a shot. Otherwise, I'll just raise as you suggested.

libraries/group_execute_provider.rb
cottsay: If it would make it more clear, we could make this `groups + [group]` as long as we reverse the iteration order on line 19.

libraries/group_execute_provider.rb
cottsay: Alright, after b73f644, you'll now be greeted with this message:
```
 Chef::Exceptions::ValidationFailed
       ----------------------------------
       Option command must be a kind of [String]!  You passed ["echo", "foo"].
```

libraries/group_execute_provider.rb
cottsay: I went ahead and reversed the iteration in 702abf3.

libraries/group_execute_provider.rb
nuclearsandwich: Great! I've never overloaded/subclassed a chef resource like this so you're blazing trails I have not.

libraries/group_execute_provider.rb
```diff
@@ -0,0 +1,28 @@
...
+      private
+
+      def command
+        command = new_resource.command
+
+        if new_resource.secondary_groups
+          groups = new_resource.secondary_groups
+          if group
+            # Always do the 'primary' group last in the call chain
+            groups = groups + [group]
```
nuclearsandwich:
```suggestion
            groups << group
```

This is mostly a style suggestion. `<<` is an append operator and aside from being more common it avoids an array allocation for `[group]`.

nuclearsandwich (approved):
A non-blocking style suggestion is my only remaining comment.

If this ends up bearing a lot of load for us we should consider trying to get secondary groups supported in the upstream execute resource.

PR #55, opened 2021-02-27
Configure jenkins-agent to use the GPG vault

cottsay:
This change drops the private keys from the jenkins-agent user and configures the user to utilize the GPG vault for signing operations.

The change to `gpg.conf` should prevent most of the GnuPG tools from automatically starting a new gpg-agent process, which should minimize the chances of accidentally overwriting the `S.gpg-agent` symbolic link with a new socket.

Requires #54

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -157,6 +157,11 @@ cookbook_file "/home/#{agent_username}/.gnupg/gpg.conf" do
   group agent_username
   mode '0600'
 end
+link "/home/#{agent_username}/.gnupg/S.gpg-agent" do
+  to '/var/run/gpg-vault/S.gpg-agent'
+  owner agent_username
+  group agent_username
+end
 
 # Set up ssh authorized keys for publish over ssh.
 directory "/home/#{agent_username}/.ssh" do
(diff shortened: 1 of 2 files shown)


recipes/repo.rb
```diff
@@ -172,38 +177,20 @@
...
-  group agent_username
-  mode '0600'
-  content gpg_key['private_key']
-end
-file "/home/#{agent_username}/.ssh/gpg_public_key.pub" do
-  owner agent_username
-  group agent_username
-  mode '0644'
-  content gpg_key['public_key']
-end
```
nuclearsandwich, line 186: We were just discussing deployment strategy offline but do you think it's worth changing these to
```ruby
file "/home/#{agent_username}/.ssh/gpg_private_key.sec" do
  action :delete
end
```

so that manual steps are not required? If we wanted to do it as a two step process we could leave these file resources un-touched for the initial deployment and then remove them in a follow-up change but I would be worried that if we just delete these resources we'll forget to clean up the old files via chef.

nuclearsandwich (approved):

recipes/repo.rb
cottsay: We could do this, but it still wouldn't be enough. We need to remove the private key from the `jenkins-agent` keyring as well. Doing that directly conflicts with the switch to use the `gpg-vault`, so special consideration would be necessary to ensure that the `gpg-vault` switch happened after the "upgrade" logic to drop the keypair files and remove the private key from the keyring.

I can pursue adding this "upgrade" logic if you like.

PR #60, opened 2021-02-27
Grant the 'pulp' user access to gpg-vault

cottsay:
There are two things about the pulp user's access that differ from how things are set up for reprepro (in the jenkins-agent user):
1. There isn't a symlink for the gpg-agent socket because we want to invoke the signing operations in docker, where we need to specifically mount the socket anyway.
2. Pulp requires that we mark the public key we're going to sign with as 'trusted'.

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -285,9 +285,9 @@ end
 
 execute 'gpg-init' do
   command 'gpg -K'
-  env 'HOME' => '/home/pulp'
   user 'pulp'
   group 'pulp'
+  environment 'HOME' => '/home/pulp'
   creates '/home/pulp/.gnupg'
 end
 cookbook_file '/home/pulp/.gnupg/gpg.conf' do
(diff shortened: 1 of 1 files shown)


recipes/repo.rb
```diff
@@ -277,6 +277,39 @@
...
+  append true
+  members ['pulp']
+  action [:manage]
+end
+
+execute 'gpg-init' do
+  command 'gpg -K'
+  env 'HOME' => '/home/pulp'
+  user 'pulp'
+  group 'pulp'
```
nuclearsandwich:
```suggestion
  user 'pulp'
  group 'pulp'
  environment 'HOME' => '/home/pulp'
```
Corrected `env` to `environment` although I would not be shocked to learn that `env` just works anyway.
The order is not significant but the following execute resources order the attributes user, group, environment so I maintain that for consistent reading.

nuclearsandwich (approved):

PR #61, opened 2021-02-27
Enable metadata signing in pulp

cottsay:
Requires #60

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -384,7 +384,20 @@ if node['ros_buildfarm']['repo']['enable_pulp_services']
     command "docker run --user 1200:1200 --rm -v /var/run/postgresql:/var/run/postgresql pulp_image pulpcore-manager reset-admin-password -p #{Shellwords.shellescape(pulp_admin_password)}"
   end
 
-  # TODO * Create gnupg directory for pulp
+  template "/home/pulp/.gnupg/pulp_sign_#{gpg_key['fingerprint']}.sh" do
+    source 'pulp/pulp_sign.sh.erb'
+    owner 'pulp'
+    group 'pulp'
+    mode '0750'
+    variables Hash[
+      gpg_key_id: gpg_key['fingerprint'],
+    ]
+  end
+  signing_service_create_cmd = "from pulpcore.app.models.content import AsciiArmoredDetachedSigningService; AsciiArmoredDetachedSigningService.objects.get_or_create(name='#{gpg_key['fingerprint']}', script='/home/pulp/.gnupg/pulp_sign_#{gpg_key['fingerprint']}.sh')"
+  execute 'pulp_create_signing_service' do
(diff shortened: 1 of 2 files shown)


recipes/repo.rb
```diff
@@ -384,7 +384,20 @@
...
+    owner 'pulp'
+    group 'pulp'
+    mode '0750'
+    variables Hash[
+      gpg_key_id: gpg_key['fingerprint'],
+    ]
+  end
+  signing_service_create_cmd = "from pulpcore.app.models.content import AsciiArmoredDetachedSigningService; AsciiArmoredDetachedSigningService.objects.get_or_create(name='#{gpg_key['fingerprint']}', script='/home/pulp/.gnupg/pulp_sign_#{gpg_key['fingerprint']}.sh')"
+  execute 'pulp_create_signing_service' do
+    command "docker run --user 1200:1200 --rm -e HOME=/home/pulp -v #{pulp_data_directory}:/var/repos/.pulp -v /var/run/postgresql:/var/run/postgresql -v /home/pulp/.gnupg:/home/pulp/.gnupg -v /var/run/gpg-vault/S.gpg-agent:/home/pulp/.gnupg/S.gpg-agent pulp_image pulpcore-manager shell --command=#{Shellwords.shellescape(signing_service_create_cmd)}"
```
nuclearsandwich, line 398:
```suggestion
    command "docker run --user 1200:1200 --rm -e HOME=/home/pulp -v #{pulp_data_directory}:/var/repos/.pulp -v /var/run/postgresql:/var/run/postgresql -v /home/pulp/.gnupg:/home/pulp/.gnupg -v /var/run/gpg-vault/S.gpg-agent:/home/pulp/.gnupg/S.gpg-agent pulp_image pulpcore-manager shell --command='#{Shellwords.shellescape(signing_service_create_cmd)}'"
```

Should the shell-escaped value to `--command=` be quoted?

cottsay: My take on this is that the point of the quotes would be to prevent spaces from breaking up the command argument. The shell escaping escapes spaces, so the quotes don't seem to be necessary any more. It might even break things - I'm not sure what escaping a space in a quoted string would do.

cottsay: I actually never tried it with the quotes present, but it does work without them.

nuclearsandwich (approved):

recipes/repo.rb
nuclearsandwich: I think the only benefit of surrounding quotes is human readability for seeing where the value to `--command=` ends via the starting and ending delimiters.

cottsay: That does make sense.
Unfortunately, a quick smoke test shows that it would indeed change the behavior (in GNU bash 5.0.17):
```
$ echo foo\ bar
foo bar
$ echo "foo\ bar"
foo\ bar
$ echo 'foo\ bar'
foo\ bar
```

PR #62, opened 2021-02-27
Address yamllint warnings

cottsay:


nuclearsandwich (approved):

.github/workflows/main.yml
```diff
@@ -1,6 +1,7 @@
 ---
 name: ci
 
+# yamllint disable-line rule:truthy
```
nuclearsandwich, line 4: What does this line do?

nuclearsandwich: Nevermind, I see it now. yamllint, whether rightly or not, is seeing the `on` field name and warning that a value of `on` would result in a boolean value. In my opinion yamllint should't complain about field names but it is so this convinces it to stop.

PR #63, opened 2021-03-03
Add upgrade path which removes previously deployed key

cottsay:
Everything is keyed off from the presence of the private key file on disk. The path is largely skipped if that file has already been removed.

I tested converging this change both with and after #55.

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -142,6 +142,34 @@ group 'gpg-vault' do
   action [:manage]
 end
 
+# Remove previous GPG deployment
+execute "gpg-agent --daemon" do
+  user agent_username
+  group agent_username
+  environment 'HOME' => "/home/#{agent_username}"
+  only_if { ::File.exist?("/home/#{agent_username}/.ssh/gpg_private_key.sec") }
+  not_if "gpg-agent"
+end
+execute "(echo y) | gpg --command-fd 0 --no-tty --yes --delete-secret-keys #{gpg_key['fingerprint']}" do
+  user agent_username
+  group agent_username
+  environment 'HOME' => "/home/#{agent_username}"
+  only_if { ::File.exist?("/home/#{agent_username}/.ssh/gpg_private_key.sec") }
+  only_if "gpg --list-secret-keys #{gpg_key['fingerprint']}"
+end
+execute "gpgconf --kill gpg-agent" do
+  user agent_username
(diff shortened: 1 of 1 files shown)


recipes/repo.rb
```diff
@@ -142,6 +142,34 @@
   action [:manage]
 end
 
+# Remove previous GPG deployment
+execute "gpg-agent --daemon" do
```
cottsay, line 146: Explicitly starting the `gpg-agent` like this will overwrite our symlink, if present. It also bypasses the `no-autostart` option in the `gpg.conf`.

recipes/repo.rb
```diff
@@ -142,6 +142,34 @@
...
+  environment 'HOME' => "/home/#{agent_username}"
+  only_if { ::File.exist?("/home/#{agent_username}/.ssh/gpg_private_key.sec") }
+  not_if "gpg-agent"
+end
+execute "(echo y) | gpg --command-fd 0 --no-tty --yes --delete-secret-keys #{gpg_key['fingerprint']}" do
+  user agent_username
+  group agent_username
+  environment 'HOME' => "/home/#{agent_username}"
+  only_if { ::File.exist?("/home/#{agent_username}/.ssh/gpg_private_key.sec") }
+  only_if "gpg --list-secret-keys #{gpg_key['fingerprint']}"
```
nuclearsandwich: Can multiple `only_if`s be used here and both work? I've not seen any documentation that says they stack or don't.

nuclearsandwich (approved):

recipes/repo.rb
cottsay: I'm actually not sure. It did what I wanted it to do, so I just assumed.

How can this be written as a compound conditional?

nuclearsandwich:
```suggestion
  only_if "test -f /home/#{agent_username}/.ssh/gpg_private_key.sec && gpg --list-secret-keys #{gpg_key['fingerprint']}"
```

I think the most straightforward way is to test file existence in shell rather than with the Ruby methods.

cottsay: > I think the most straightforward way is to test file existence in shell rather than with the Ruby methods.

Ah, good call.

cottsay:
Alright, I ran through a few tests after committing the suggestion, and things appear to be in order. Merging.

PR #65, opened 2021-03-04
Don't proxy nonexistent pulp services

cottsay:
If pulp services are disabled, even if RPM repositories are configured, we shouldn't create the proxy entries.

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -555,7 +555,7 @@ package 'nginx'
 template '/etc/nginx/sites-available/repo' do
   source 'nginx/repo.conf.erb'
   variables Hash[
-    rpm_repos: node['ros_buildfarm']['rpm_repos']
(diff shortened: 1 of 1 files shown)


recipes/repo.rb
```diff
@@ -555,7 +555,7 @@
 template '/etc/nginx/sites-available/repo' do
   source 'nginx/repo.conf.erb'
   variables Hash[
-    rpm_repos: node['ros_buildfarm']['rpm_repos']
+    rpm_repos: (node['ros_buildfarm']['repo']['enable_pulp_services'] ? node['ros_buildfarm']['rpm_repos'] : [])
```
nuclearsandwich, line 558: Now that we have the RPM repos attribute I think it makes sense to refactor and make the presence of RPM repositories the condition for whether pulp is enabled rather than relying on a separate boolean. I can turn this PR into that if you agree.

nuclearsandwich (changes requested):

cottsay:

PR #67, opened 2021-03-04
Initialize empty RPM repositories

cottsay:
Since `createrepo` nor `createrepo_c` are available for Focal, I generated some empty metadata that we can copy into each new RPM repository.

This change also creates the directory structure for the repositories. When Pulp is enabled, it should proxy the leaf directories for the building repository, but we still want the directory structure leading up to those leaves. For simplicity, and to better support a possible future where pulp isn't used, I'm still creating metadata for the building repos even though it should never actually be exposed.

I'm also creating an `RPM-GPG-KEY-foo` file for each repo (building/testing/main). The typical pattern would be to include a `foo.repo` file next to the GPG key, but I need to embed the absolute URL to the repository there, so we'll need to make that available/configurable somehow.


recipes/repo.rb
```diff
@@ -551,6 +551,79 @@
   end
 end
 
+node['ros_buildfarm']['rpm_repos'].each do |dist, versions|
```
nuclearsandwich: Is there a reason this is being done in a second iteration rather than integrating it into the iteration above?

nuclearsandwich (approved):
Looks good aside from a question about why it runs in a second iteration rather than alongisde the pulp initialization.

recipes/repo.rb
cottsay: The idea was to keep the Pulp-specific stuff separate from the general RPM stuff. If we want to just go all-in on Pulp, I can refactor this.

nuclearsandwich: Got it. I see the logic there but I don't think we have a ready alternative and I think that for now rpms == pulp is a fine expectation.

cottsay: Unfortunately the iteration doesn't line up because pulp views the three sub-repos as part of the same Pulp repo, while the directory structure doesn't reflect that. I did move it into the first level of the iteration loop in 067fb2a, though.

nuclearsandwich: At the risk of seeming mercurial do you find the first or second commit more readable? If we can't share the single compound iteration I think that readability becomes the next highest priority. I'm good either way with a slight preference toward the current approach but yield the final decision if you want to restore the first commit with the separate iteration.

nuclearsandwich (approved):
Left my thoughts on the second commit and am good leaving the decision to the implementer. Thanks for iterating!

PR #69, opened 2021-03-05
Drop explicit enable flag for Pulp

cottsay:
Instead, use the presence of RPM repository configurations to decide whether to deploy pulp services.

Closes #65

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -386,7 +386,7 @@ cookbook_file "#{pulp_data_directory}/Dockerfile" do
   source 'pulp/Dockerfile'
 end
 
-if node['ros_buildfarm']['rpm_repos']
+if node['ros_buildfarm']['rpm_repos'] and not node['ros_buildfarm']['rpm_repos'].empty?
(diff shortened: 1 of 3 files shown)


recipes/repo.rb
```diff
@@ -386,7 +386,7 @@
   source 'pulp/Dockerfile'
 end
 
-if node['ros_buildfarm']['repo']['enable_pulp_services']
+if node['ros_buildfarm']['rpm_repos']
```
nuclearsandwich: Unlike Python where empty collections are false-y all objects in Ruby are truth-y except false and nil.

```suggestion
if not node['ros_buildfarm']['rpm_repos'].empty?
```

nuclearsandwich (approved):
There's one implementation issue that must be resolved before merging but otherwise looking good.

recipes/repo.rb
cottsay: Thanks, I expanded on this in 457f131.

cottsay:

PR #70, opened 2021-03-05
Restart jenkins-agent immediately on unit update

cottsay:
It appears that allowing the delayed restart to run immediately after the service was started is causing it to fail occasionally.

This change issues a restart immediately after the unit file is created or updated (and after systemd reloads it). In the case where the unit was updated, the explicit 'start' is therefore a no-op. Otherwise, the service is started at that time.

### DIFF
--- a/recipes/agent.rb
+++ b/recipes/agent.rb
@@ -139,7 +139,7 @@ template '/etc/systemd/system/jenkins-agent.service' do
     username: agent_username,
   ]
   notifies :run, 'execute[systemctl-daemon-reload]', :immediately
-  notifies :restart, 'service[jenkins-agent]'
+  notifies :restart, 'service[jenkins-agent]', :immediately
 end
 
 execute 'systemctl-daemon-reload' do


nuclearsandwich:
> It appears that allowing the delayed restart to run immediately after the service was started is causing it to fail occasionally.

Do you have an example for the record?

cottsay:
> Do you have an example for the record?

https://github.com/ros-infrastructure/cookbook-ros-buildfarm/runs/2042717582

```
Mar 05 20:17:03 dokken systemd[1]: jenkins-agent.service: Start request repeated too quickly.
Mar 05 20:17:03 dokken systemd[1]: jenkins-agent.service: Failed with result 'start-limit-hit'.
Mar 05 20:17:03 dokken systemd[1]: Failed to start ROS build farm jenkins agent.
```

cottsay:
This one just hit it, too: https://github.com/ros-infrastructure/cookbook-ros-buildfarm/runs/2042934273
And another: https://github.com/ros-infrastructure/cookbook-ros-buildfarm/runs/2042987550

cottsay:
It might be a little more complicated than I thought. Looks like part of the reason the rate limit triggered was because of automatic restarts trying to get the service running after some other failure.

That should probably be looked after as well, but not starting and immediately restarting the service would certainly mitigate the issue.

cottsay:
Well, now the re-run on this change failed too. Clearly not as effective at mitigating the issue as I thought.

PR #72, opened 2021-03-05
Revert "Use xslt to customize autoindex pages."

nuclearsandwich:
Reverts ros-infrastructure/cookbook-ros-buildfarm#49

This is causing problems when interacting with the repodata.xml from Pulp. I need to figure out how best to make sure that this only gets used on the auto-index pages rather than on hosted XML content.


cottsay (approved):
The file will be left on disk in the staging environment, but I think that's OK for now.

nuclearsandwich:
The one failed job is due to an error reaching pypi.org and since everything else passed I'm going to go ahead and merge.

PR #73, opened 2021-03-06
Add credentials-binding plugin for Jenkins

cottsay:
I do need some guidance on this one for how to confirm that the plugin's dependencies are met by the currently declared set of plugins in this list.

### DIFF
--- a/attributes/plugins.rb
+++ b/attributes/plugins.rb
@@ -76,7 +76,7 @@ default['ros_buildfarm']['jenkins']['plugins'] = {
   "pam-auth" => "1.5.1",
   "parameterized-trigger" => "2.35.2",
   "pipeline-utility-steps" => "2.5.0",
-  "plain-credentials" => "1.4",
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
> how to confirm that the plugin's dependencies are met by the currently declared set of plugins in this list.

Generally to sort out dependencies in the past we've installed the plugin on a host and then dumped the plugin versions to this list. You can also check them on https://plugins.jenkins.io/credentials-binding/#dependencies 

From comparing the dependencies to the current list of plugins it seems like we will need to perform a few updates to the other installed plugins.

cottsay:
Thanks, @nuclearsandwich. I bumped the necessary deps in b5b2026.
* https://plugins.jenkins.io/plain-credentials/#releases
* https://plugins.jenkins.io/workflow-step-api/#releases

nuclearsandwich (approved):
It's probably worth doing a plugin update sweep soon but no need to block this change.

PR #75, opened 2021-03-09
Add optional username field to Jenkins password credentials.

nuclearsandwich:
Since pulp needs one I used the pretend pulp cred as the test data bag for this.


cottsay (approved):

nuclearsandwich:
The CI failure is the flaky jenkins-agent start that we've been seeing and still needs investigation but it's not related to this change.

PR #76, opened 2021-03-09
Add pulp_base_url credential

cottsay:


nuclearsandwich (approved):
Strictly speaking I'm not sure a test credential is needed here but it's a good reminder until proper documentation for data_bags are written and available.

PR #79, opened 2021-03-16
Proxy all 3 repositories to Pulp, not just 'building'

cottsay:
Eventually we'd like to manifest the repositories as files-on-disk so that they can be rsync'd to mirrors, but for now, this is needed to unblock testing.


nuclearsandwich (approved):
Looks correct to me.

PR #81, opened 2021-04-06
Add pulp_rsync content endpoint

cottsay:
If you squint, it almost looks like an rsync implementation: https://github.com/cottsay/pulp_rsync

It's close enough for what we need it for.

To be clear, the idea is to use rsync to hydrate the local filesystem. I'm not planning to expose `pulp_rsync` to the interwebs.

### DIFF
--- a/files/pulp/Dockerfile
+++ b/files/pulp/Dockerfile
@@ -15,6 +15,7 @@ RUN dnf install \
   expat-devel \
   file-devel \
   gcc \
+  git \
   glib2-devel \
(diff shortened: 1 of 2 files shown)


nuclearsandwich (approved):

nuclearsandwich:
I didn't review the rsync implementation itself and instead chose to squint.

cottsay:
> I didn't review the rsync implementation itself and instead chose to squint.

Wise.

PR #83, opened 2021-04-07
Add support for an RPM bootstrap repository

cottsay:
In `ros_buildfarm`, an `import-upstream` job for RPMs is implemented by searching for repos with a certain name (i.e. ros-buildfarm-foo-bar-baz) and triggering a 'sync' operation on those repos, then performing an operation similar to the sync operation that we do between building/testing/main, but between the candidate repo and each of building, testing, and main.

Creating the repo that has the 'sync' operation is similar to creating any other repo, except that it has a 'remote' associated with it.

This change implements the provisioning steps for the remote and the associated repo.

### DIFF
--- a/attributes/repo.rb
+++ b/attributes/repo.rb
@@ -38,7 +38,7 @@ default['ros_buildfarm']['repo']['container_registry_cache_enabled'] = true
 default['ros_buildfarm']['repo']['pulp_worker_count'] = 2
 default['ros_buildfarm']['repo']['enable_pulp_services'] = true
 default['ros_buildfarm']['rpm_repos']['rhel']['8'] = %w[x86_64]
-default['ros_buildfarm']['rpm_bootstrap_repos']['rhel']['8'] = Hash[
+default['ros_buildfarm']['rpm_upstream_repos']['bootstrap']['rhel']['8'] = Hash[
   architectures: %w[x86_64],
   binary: 'http://repos.ros.org/repos/rhel/ros_bootstrap/8/$basearch/',
   debug: 'http://repos.ros.org/repos/rhel/ros_bootstrap/8/$basearch/debug/',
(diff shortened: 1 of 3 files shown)


nuclearsandwich (approved):
My main hesitation with using chef attributes for bootstrap repository configuration is that it means that we (and probably more likely community buildfarms) can't have multiple import sources provided by multiple bootstrap repository config files.

We're talking with one project which is investigating how to use the ROS upstream repositories as an import source and then running their buildfarm on solely their project's packages. While the repository import is transitive (that is, everything in the bootstrap repository will be pulled into the ROS repositories that they import from) but there may also be buildfarms which inject proprietary non-ROS packages via the import_upstream job with a custom config.

That concern aside, I think the rest of this change is solid and I think my preference is to take this as-is unless you think that "graduating" to a runtime configurable bootstrap repository is going to be more challenging than delaying this PR to incorporate it from the first.

cottsay:
Let's discuss this a little more. I think some slight tweaks will make this easy to do.

I mentioned in the description that a pattern match is used to discover the 'upstream' repositories to sync. That means that if multiple exist, multiple will sync. So the way this is set up, I think it already meets your criteria for runtime-configurable. The tweak that we need to make is to set the default pattern to be more inclusive.

Here's where the pattern is defined in `ros_buildfarm`: https://github.com/ros-infrastructure/ros_buildfarm/blob/4240ad951a89ef9ebae1fac87f1f42423178e709/ros_buildfarm/templates/release/rpm/import_upstream_job.xml.em#L24-L29

What I'm proposing is to look for `ros-upstream-.*-rhel-8-x86_64` instead of `ros-bootstrap-rhel-8-x86_64`, and then make the default provisioning done here create an entry for `ros-upstream-bootstrap-rhel-8-x86_64`.

For provisioning additional upstream targets, we could either change the attribute structure to allow for multiple repository definitions, or we could document how to use `rpm_repo_init.py` to do it directly. Either way, the provisioning presented here won't remove existing entities from pulp - it only creates the pulp entities if they don't exist. However they get configured, if multiple are present, they'll all get processed.

nuclearsandwich (approved):
Thanks for iterating. I think that the extra layer nets us a lot of flexibility when setting up more complex imports for customized buildfarm deployments.

PR #88, opened 2021-04-14
Enable caching of some of the RPM repository metadata

cottsay:
The names of files which match the caching rule contain the content hashes, so the content at that URL should never change. These files disappear as the repository is updated, and this leads to a race between the loading of the repomd.xml and accessing the files listed within (and here in this config).

This isn't a complete solution for a variety of reasons, among them that it depends on a client accessing the content so that it can be cached to begin with. I'm hoping it will mitigate the problem, however.

### DIFF
--- a/templates/nginx/repo.conf.erb
+++ b/templates/nginx/repo.conf.erb
@@ -1,3 +1,5 @@
+proxy_cache_path /var/cache/nginx/rpm_md_cache levels=1:2 keys_zone=rpm_md_cache:10m max_size=1g inactive=1d use_temp_path=off;
+
 server {
 	listen 80 default_server;
 	server_name repo;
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
>  it depends on a client accessing the content so that it can be cached to begin with. I'm hoping it will mitigate the problem, however.

If this proves an incomplete solution we could consider warming the caches as a finishing step for jobs that update repository data. I'm not sure how much time that would add to import_package jobs though.

nuclearsandwich:
CI is definitely in a troubled state at the moment but the errors on the repo CI are nginx related and I expect are due to issues with the generated config file.

cottsay:
> the errors on the repo CI are nginx related

Good catch - I made a late change to the cache directory that broke things. It is resolved now.

nuclearsandwich (approved):

cottsay:

PR #91, opened 2021-04-23
Update plugins to address security advisories.

nuclearsandwich:
Plugins removed
---------------
* analysis-core
    This plugin was actually removed earlier in PR #23 but it will still
    be present on systems which were initially provisioned before the
    plugin was yanked.
* config-file-provider
    I'm the least sure about this one but as far as I can tell it is not
    a dependency of any used plugin and is not used by any jobs in
    ros_buildfarm. Plugin is up for adoption but does have an updated
    version with fixes for the security issues if it turns out we need
    it.
* cvs
    ros_buildfarm (assessed via grep) does not have any support for cvs
    repositories. The CVS plugin was previously installed by default on
    new Jenkins installations and that may be how it ended up here.
* icon-shim
    This plugin was added to backport functionality to earlier Jenkins
    instances (<= 1.576) and the plugin library itself has been merged
    into Jenkins 2.269. It can now be unambiguously safely removed.

Plugins updated
---------------
* configuration-as-code 1.43 -> 1.47
* credentials 2.3.11 -> 2.3.18
* extra-colums 1.20 -> 1.23
* matrix-auth 2.6.5 -> 2.6.6
* mercurial 2.4 -> 2.14
* scm-api 2.6.3 -> 2.6.4
* structs 1.20 -> 1.22
* subversion 2.13.1 -> 2.14

Changelogs for updated packages were reviewed and found to be benign.
Only packages with pending security advisories were updated.

Several additional plugin security advisories are outstanding but will require more effort to upgrade as there are incompatible changes.


attributes/remove_plugins.rb
```diff
@@ -0,0 +1,9 @@
+# Plugins on this list will be removed on each chef run.
+# Adding plugins to this list requires a minor version bump.
+# Removing plugins from this list requires a major version bump.
+default['ros_buildfarm']['jenkins']['remove_plugins'] = %w[
+analysis-core
+config-file-provider
+cvs
+icon-shim
```
cottsay: It would be nice to match the indentation used in `plugins.rb`. I understand that it isn't the same data type, though.
```suggestion
  analysis-core
  config-file-provider
  cvs
  icon-shim
```

cottsay (approved):

attributes/remove_plugins.rb
nuclearsandwich: It still makes sense to do a visual indent within the array though. Good call :+1:

nuclearsandwich:
I just remembered that I have one more set coming because I also need to update subversion (and it will touch structs, credentials, and some others) but plugins.jenkins.io was down so I got stalled and then forgot about it.

PR #92, opened 2021-04-23
Use systemd to manage GPG vault socket directory

cottsay:
The directory currently being created by Chef doesn't persist across reboots. We can use systemd to manage the directory instead, which should mean that it is always created prior to starting, and always deleted after stopping.

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=

Additionally, ensure that systemd reloads the gpg-vault-agent unit file when it is changed.

### DIFF
--- a/files/gpg-vault-agent.service
+++ b/files/gpg-vault-agent.service
@@ -5,8 +5,10 @@ Documentation=man:gpg-agent(1)
 [Service]
 Type=forking
 User=gpg-vault
+RuntimeDirectory=gpg-vault
+RuntimeDirectoryMode=750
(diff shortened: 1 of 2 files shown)


nuclearsandwich (approved):
https://docs.chef.io/resources/systemd_unit/ handles the inner workings of daemon-reload and friends for us but I tried it somewhere (forgetting exactly where) and I found it less ergonomic than the template-for-unit-file and service approach even with the daemon-reload step. If we start to get encumbered by the daemon-reload biz it may be worth a fresh look in time.

nuclearsandwich:
Agent CI errors are known flaky tests. Flakes will eat the software world :sob:

PR #95, opened 2021-05-25
Add certificate update hook script for acme.sh.

nuclearsandwich:
Nginx needs to be reloaded or restarted in order to re-read SSL
certificate files that acme.sh has updated.

In case other services ever need to be similarly restarted a separate
hook script is used to remove the need for updating the reloadcmd
directly.

This will resolve the issue we have periodically where acme.sh updates the certificate but nginx must be manually restarted before the new certificate will be served.

### DIFF
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -257,8 +257,12 @@ if node['ros_buildfarm']['letsencrypt_enabled']
       --fullchain-file #{cert_path}
       --key-file #{key_path}
       --reloadcmd /root/cert-update-hook.sh
+      --force
     )
-    not_if "test -d /root/.acme.sh/#{server_name}"
+    not_if {
+      File.directory?("/root/.acme.sh/#{server_name}") and not
+      File.read("/root/.acme.sh/#{server_name}/#{server_name}.conf").match(/Le_ReloadCmd=''/)
+    }
   end
 else
   template '/etc/nginx/sites-enabled/jenkins' do
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -262,7 +262,7 @@ if node['ros_buildfarm']['letsencrypt_enabled']
     not_if {
       # TODO the second guard clause can be removed after >= 0.6.0
       File.directory?("/root/.acme.sh/#{server_name}") and
-      File.read("/root/.acme.sh/#{server_name}/#{server_name}.conf").match(/Le_ReloadCmd='__ACME_BASE64__START_c3lzdGVtY3RsIHJlc3RhcnQgbmdpbng=__ACME_BASE64__END_'/)
+      File.read("/root/.acme.sh/#{server_name}/#{server_name}.conf").match(/Le_ReloadCmd='__ACME_BASE64__START_L3Jvb3QvY2VydC11cGRhdGUtaG9vay5zaA==__ACME_BASE64__END_'/)
     }
   end
 else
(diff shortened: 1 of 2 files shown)


recipes/jenkins.rb
```diff
@@ -249,8 +256,14 @@ def jenkins = Jenkins.getInstance()
...
       --fullchain-file #{cert_path}
       --key-file #{key_path}
+      --reloadcmd /root/cert-update-hook.sh
+      --force
     )
-    not_if "test -d /root/.acme.sh/#{server_name}"
+    not_if {
+      # TODO the second guard clause can be removed after >= 0.6.0
+      File.directory?("/root/.acme.sh/#{server_name}") and
+      File.read("/root/.acme.sh/#{server_name}/#{server_name}.conf").match(/Le_ReloadCmd='__ACME_BASE64__START_c3lzdGVtY3RsIHJlc3RhcnQgbmdpbng=__ACME_BASE64__END_'/)
```
nuclearsandwich: In order to allow arbitrary script fields, acme.sh base64 encodes script lines. This is the acme base64 encoding of `/root/cert-update-hook.sh`

recipes/jenkins.rb
```diff
@@ -249,8 +256,14 @@ def jenkins = Jenkins.getInstance()
       --domain #{server_name}
       --fullchain-file #{cert_path}
       --key-file #{key_path}
+      --reloadcmd /root/cert-update-hook.sh
+      --force
```
nuclearsandwich, line 260: The `--force` flag needs to be added in order to force a change to a deployed, up-to-date certificate.

recipes/jenkins.rb
```diff
@@ -249,8 +256,14 @@ def jenkins = Jenkins.getInstance()
...
     )
-    not_if "test -d /root/.acme.sh/#{server_name}"
+    not_if {
+      # TODO the second guard clause can be removed after >= 0.6.0
+      File.directory?("/root/.acme.sh/#{server_name}") and
+      File.read("/root/.acme.sh/#{server_name}/#{server_name}.conf").match(/Le_ReloadCmd='__ACME_BASE64__START_L3Jvb3QvY2VydC11cGRhdGUtaG9vay5zaA==__ACME_BASE64__END_'/)
```
nuclearsandwich, line 265: In order to allow arbitrary script fields, acme.sh base64 encodes script lines. This is the acme base64 encoding of `/root/cert-update-hook.sh`

recipes/jenkins.rb
```diff
@@ -249,8 +256,14 @@ def jenkins = Jenkins.getInstance()
       --domain #{server_name}
       --fullchain-file #{cert_path}
       --key-file #{key_path}
+      --reloadcmd /root/cert-update-hook.sh
```
j-rivero, line 259: A side of opening the door to execute more commands on reloading, why not using `--reloadcmd     "service nginx force-reload"`?

nuclearsandwich: > A side of opening the door to execute more commands on reloading, why not using `--reloadcmd "service nginx force-reload"`?

I created the hook script intentionally to add an indirection layer allowing us to expand the restarted services (in fact this entire thing is self-plagarized/adapted from my the itamae scripts I use to restart varnish, hitch, and nginx when SSL certs are renewed in my homelab).

To change our acme.sh configuration we either have to use the `--force` flag which forces an unnecessary certificate renewal or grok and modify the acme.sh configuration directly. Infrequent forced renewals won't be a problem but they are rate limited and that's an issue we could run into in the future. I also want to avoid depending on the acme.sh internal configuration to make changes.

j-rivero (approved):

PR #96, opened 2021-06-03
Notify `gpg --import` resources when key files change.

nuclearsandwich:
This patch improves configuration behavior when key contents change.
When deploying changes to the public key, the keys on disk were updated
but the gpg keyring did not import the new keys since existing keys were
present.

Now, changes to the file resources will trigger runs of the import
execute resources which will provide the same behavior as before for
a freshly provisioned host but will correctly re-import keys when those
files change.

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -117,24 +117,26 @@ file '/home/gpg-vault/.gnupg/gpg_public_key.pub' do
   owner 'gpg-vault'
   group 'gpg-vault'
   mode '0644'
+  notifies :run, 'execute[gpg --import /home/gpg-vault/.gnupg/gpg_public_key.pub]', :immediately
 end
 execute 'gpg --import /home/gpg-vault/.gnupg/gpg_public_key.pub' do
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
CI failure is exposing a failed GPG import so I must have missed something here.

nuclearsandwich:
> CI failure is exposing a failed GPG import so I must have missed something here.

Okay I've figured out what's happening. There are two identically named execute resources, one for the Jenkins agent and one for the pulp user, and the `notifies` added to the `file[/var/repos/repos.key]` resources is triggering both of them but the latter is failing because the pulp user and group does not exist yet.

Naively I think there are two ways out of this, moving the pulp user creation resoures above all gpg related file resources, or changing the pulp import and trust resources to target their own copy of the file on disk which is changed later in the recipe when the prerequisite resources have already converged.

nuclearsandwich:
I think this will be revisit-able once pulp is removed.

nuclearsandwich:
This is now updated and ready for review with pulp removed.

PR #100, opened 2021-11-18
Bump to pulp_rsync 0.0.3

cottsay:
A dropped database connection during the app's heartbeat was causing the heartbeat loop to terminate, so I added a try/except to keep the loop going.


j-rivero (approved):
Failure in CI for Chef 15 is expected:
> 2021-11-23T22:32:40.6762089Z            * remote_file[/opt/kitchen/cache/https___pkg_jenkins_io_debian-stable_jenkins_io_key] action create[2021-11-23T22:32:40+00:00] ERROR: SSL Validation failure connecting to host: pkg.jenkins.io - SSL_connect returned=1 errno=0 state=error: certificate verify failed (certificate has expired)

Chef15 has its own embedded copy of an outdated root cert.

PR #101, opened 2021-11-23
Update CI: stop using Chef 15, start using 17

j-rivero:
Chef15 is problematic since it ships an outdated root certificate making some certs to fail, see https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/100.

I'm setting 17 here to check if it is valid or we need to modify code.


nuclearsandwich:
Superceded by #105 but only because I forgot this was here. Thanks @j-rivero !

PR #104, opened 2022-01-19
Add an SSH publish endpoint for CI archives

cottsay:
This change adds an SSH publisher endpoint for storing CI archives on repo.ros2.org. As it stands, we have no plans to sync these archives off from the repo host, but that may change in the future.

For our immediate needs, getting the archive out of Jenkins should be good enough. We can re-evaluate later if bandwidth becomes a problem.

### DIFF
--- a/attributes/ssh_publisher.rb
+++ b/attributes/ssh_publisher.rb
@@ -37,4 +37,5 @@ default['ros_buildfarm']['ssh_publisher']['repo_root_dir'] = '/var/repos/ubuntu/
 default['ros_buildfarm']['ssh_publisher']['docs_root_dir'] = '/var/repos/docs'
 default['ros_buildfarm']['ssh_publisher']['rosdistro_cache_root_dir'] = '/var/repos/rosdistro_cache'
 default['ros_buildfarm']['ssh_publisher']['status_page_root_dir'] = '/var/repos/status_page'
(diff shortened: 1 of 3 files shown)


attributes/ssh_publisher.rb
```diff
@@ -37,4 +37,5 @@
 default['ros_buildfarm']['ssh_publisher']['docs_root_dir'] = '/var/repos/docs'
 default['ros_buildfarm']['ssh_publisher']['rosdistro_cache_root_dir'] = '/var/repos/rosdistro_cache'
 default['ros_buildfarm']['ssh_publisher']['status_page_root_dir'] = '/var/repos/status_page'
+default['ros_buildfarm']['ssh_publisher']['ci_archves_root_dir'] = '/var/repos/ci'
```
nuclearsandwich:
```suggestion
default['ros_buildfarm']['ssh_publisher']['ci_archves_root_dir'] = '/var/repos/ci_archives'
```

Any reason why the path name shouldn' tmatch the configuration name?

nuclearsandwich (approved):
A small nitpick but otherwise looks good to me.

attributes/ssh_publisher.rb
cottsay: 560e728

PR #105, opened 2022-01-27
Update project to cinc/chef 17

nuclearsandwich:
Chef 15 is no longer supported and issues with OpenSSL certificates are blocking deployment using it. 
I was locally testing the move first to Chef 16 and when that went fine I bumped to 17 to see the damage.
There was only one deprecation warning which was resolved with an update to the latest version of the Jenkins cookbook.
That update also simplifies the dependency situation (resolving the deprecation warning) and so is generally positive.

The Jenkins role is still unable to converge due to a plugin being delisted which is being handled in a separate PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,8 +26,7 @@ jobs:
           - 'repo'
           - 'jenkins'
         chef_version:
-          - '15'
-          - '16'
+          - '17'
       fail-fast: false
     steps:
       - name: Check out code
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
CI from https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/106 can be considered for this review since it includes these changes as well as mitigations for another issue.

nuclearsandwich:
Merging in order to deploy to test farm.

PR #107, opened 2022-02-02
Update structs plugin version.

nuclearsandwich:
Version >= 1.23 is required for the bundled publish-over-ssh plugin.

The latest version is 308.v852b473a2b8c which is of odd enough
construction that it looks like it could be a misconfigured auto
release so I'm avoiding it for now.


nuclearsandwich:
Merging for deployment.

PR #109, opened 2022-02-16
Update seccomp profile to moby/moby 20.10.12.

nuclearsandwich:
This pulls in the default seccomp profile from moby/moby 20.10.12 which includes, among other changes, support for the clone3 system call which is being blocked in our current Ubuntu Jammy sourcedeb containers.

The personality system call for sbcl in ROS 1 is maintained.

### DIFF
--- a/files/docker-seccomp-default-with-personality.json
+++ b/files/docker-seccomp-default-with-personality.json
@@ -74,6 +74,7 @@
 				"clock_nanosleep",
 				"clock_nanosleep_time64",
 				"close",
+				"close_range",
 				"connect",
 				"copy_file_range",
 				"creat",
@@ -85,6 +86,7 @@
 				"epoll_ctl",
 				"epoll_ctl_old",
 				"epoll_pwait",
+				"epoll_pwait2",
 				"epoll_wait",
 				"epoll_wait_old",
 				"eventfd",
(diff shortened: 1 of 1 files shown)


ros-discourse:

https://discourse.ros.org/t/holding-releases-to-perform-rolling-transition-to-new-platforms/24191/12

cottsay (approved):

files/docker-seccomp-default-with-personality.json
```diff
@@ -815,4 +840,4 @@
 			"excludes": {}
 		}
 	]
-}
+}
```
nuclearsandwich, line 843: Since it came up in an offline discussion. The newline change here is present in upstream which is why it's reflected in this diff.

 I actually cannot recall whether I initially created this file by copy-pasting and thus the final newline was added by my editor's (very appropriate) general policy of always having one or whether this is a change that occurred upstream as well but this file was fetched directly from the upstream tree via GitHub with no final newline.

j-rivero:
Seems good to my eyes:
```diff
cookbook-ros-buildfarm/files on  update-seccomp-policy [?] ❯ diff -u /tmp/moby-20.10.12/profiles/seccomp/default.json docker-seccomp-default-with-personality.json 
--- /tmp/moby-20.10.12/profiles/seccomp/default.json	2021-12-12 11:11:51.000000000 +0100
+++ docker-seccomp-default-with-personality.json	2022-02-16 21:09:30.344735175 +0100
@@ -234,6 +234,7 @@
 				"openat",
 				"openat2",
 				"pause",
+				"personality",
 				"pidfd_open",
 				"pidfd_send_signal",
 				"pipe",
```

PR #110, opened 2022-04-15
Add letsencrypt support for the repository host.

nuclearsandwich:
Like the Jenkins host, the repo host may also be publicly exposed. Apt repositories don't require https for repository integrity since that's achieved through GPG key signing repository metadata.

However other resources on the repository host may benefit from having a secure transport mechanism. Including the secure distribution of the public key information used for the apt security process or other artifacts hosted on the repository server.

### DIFF
--- a/attributes/repo.rb
+++ b/attributes/repo.rb
@@ -21,6 +21,10 @@ default['ros_buildfarm']['apt_repos']['suites'] = %w[xenial bionic focal stretch
 # ```
 default['ros_buildfarm']['repo']['rsyncd_endpoints'] = Hash[]
 
+# The repository host server name is currently only used to provide
+# a server name when using letsencrypt for https and is environment-specific.
+default['ros_buildfarm']['repo']['server_name'] = nil
+
 # The ros_buildfarm jobs are configured to try and pull the latest container
 # image from the Docker registry at the start of each job.  To prevent that
 # from consuming excess bandwidth and rate limited Docker API requests the repo
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -689,12 +689,53 @@ if node['ros_buildfarm']['repo']['enable_pulp_services']
 end
 
 package 'nginx'
+service 'nginx' do
+  action [:start, :enable]
+end
+
+
+server_name = node['ros_buildfarm']['repo']['server_name']
+cert_path = if server_name
+              "/etc/ssl/certs/#{server_name}/fullchain.pem"
+            else
+              nil
+            end
+key_path = if server_name
+             "/etc/ssl/private/#{server_name}.key"
+           else
+             nil
+           end
+
+# Initial configuration of SSL certs for HTTPS
+if node['ros_buildfarm']['letsencrypt_enabled']
+  if server_name.nil?
+    Chef::Log.fatal "No repo server name set for #{node.chef_environment} but it is required when using letsencrypt."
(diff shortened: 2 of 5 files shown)


attributes/repo.rb
```diff
@@ -21,6 +21,10 @@
 # ```
 default['ros_buildfarm']['repo']['rsyncd_endpoints'] = Hash[]
 
+# The repository host server name is currently only used to provide
+# a server name when using letsencrypt for https and is environment-specific.
+default['ros_buildfarm']['repo']['server_name'] = nil
```
nuclearsandwich, line 26: I am bit conflicted about defaulting this to `nil`, the jenkins server
name defaults to `'ros_buildfarm'` and would biff if used with
letsencrypt enabled since it isn't a valid fqdn for the server name
verification.

`nil` breaks the convention of having a default but also gives us the
ability to raise the alarm early if a configuration isn't going to work,
rather than having the acme.sh certificate issue attempt fail.

recipes/repo.rb
```diff
@@ -689,12 +689,53 @@
...
   variables Hash[
-    rpm_repos: node['ros_buildfarm']['rpm_repos']
+    letsencrypt_enabled: node['ros_buildfarm']['letsencrypt_enabled'],
+    cert_path: cert_path,
+    key_path: key_path,
+    rpm_repos: node['ros_buildfarm']['rpm_repos'],
+    server_name: server_name,
   ]
-  notifies :restart, 'service[nginx]'
+  notifies :restart, 'service[nginx]', :immediately
```
nuclearsandwich, line 738: We need this configuration to be applied before the acme certificate issuance is triggered below.

templates/nginx/repo.conf.erb
```diff
@@ -1,8 +1,20 @@
 server {
 	listen 80 default_server;
-	server_name repo;
+	<% if @letsencrypt_enabled %>
+	listen 443 ssl default_server;
```
nuclearsandwich, line 4: `default_server` here is a per-address and per-port setting so these two do not conflict with each other.

The directive makes it so that even if someone is just requesting the server by IP address this site will be served for HTTP (80) or HTTPS (443) even without knowing the requested server name.

recipes/repo.rb
```diff
@@ -689,12 +689,53 @@
...
 end
```
nuclearsandwich, line 739: Unlike the Jenkins configuration, we make all the same content available via HTTP or HTTPS instead of redirecting to HTTPS.

Thus it makes more sense to use a single template with some variable options than it does to use separate templates as we do for the Jenkins host nginx config.

nuclearsandwich:
The yamllint check appears to be going through some strife upstream at the moment. I'm content to ignore it for a little until it settles down. I'm more perturbed by the Jenkins agent failure which looks like it could be a configuration race with authentication. I'll investigate that for sure.

cottsay (approved):
Thanks for the review guides. This looks great.

nuclearsandwich:
I can also reproduce the jenkins cookbook CI failure on `latest`. So I'm going to make a case that we override the failed check and merge this since the recipe that we primarily changed is passing.

That being said, I _did_ move some stuff around in the jenkins recipe so it's possible that I've introduced a regression hiding behind the current failure but I'm willing to assume that risk and address it in a follow-up PR after I've fixed the current issue on another branch.

PR #113, opened 2022-06-08
Add createrepo-agent to repo host alongside Pulp

cottsay:
This change adds the necessary configuration for using createrepo-agent on the repo host for RPM repository management, and creates a separate repository tree alongside the Pulp one for comparison.

When we're ready to switch, just drop the `_cra` suffix from the repository directory name and remove all of the pulp stuff.

### DIFF
--- a/recipes/repo.rb
+++ b/recipes/repo.rb
@@ -321,6 +321,13 @@ node['ros_buildfarm']['rpm_repos'].each do |dist, versions|
         not_if { ::File.exist?("#{srpms_dir}/repodata/repomd.xml") }
       end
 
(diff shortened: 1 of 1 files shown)


cottsay:
Is there an official branch we're using for the staging farm that I should redirect this PR into?

nuclearsandwich (approved):
I gave some brief consideration to asking for these additions to be scoped to the presence of rpm repos in the config. But the tools are light enough and unlike pulp, don't utilize resources when they aren't being used that it makes more sense to make them an unconditional default.

nuclearsandwich:
@cottsay I think this was superseded by #116. Can it be closed?

cottsay:
> @cottsay I think this was superseded by #116. Can it be closed?

Yes.

PR #116, opened 2022-08-25
Switch from Pulp to createrepo-agent for RPMs

cottsay:
This change switches the RPM repository management solution from Pulp to `createrepo-agent`. It leaves Pulp configured and running, but moves the endpoint from `http://example.com/{os_name}/building/{os_verison}/{os_arch}` to `http://example.com/{os_name}_pulp/building/{os_verison}/{os_arch}`. Existing `testing` and `main` repositories are left untouched, and will be managed by `createrepo-agent` automatically on-demand.

There is a one-time manual migration that must be performed to copy the existing contents of the `building` repository from Pulp to `createrepo-agent`. After deployment is complete, run the following command on the repo host **as the `jenkins-agent` user**:
```
$ createrepo-agent --sync=http://127.0.0.1/rhel_pulp/building/8/ --arch=SRPMS --arch=x86_64 /var/repos/rhel/building/8/
```

### DIFF
--- a/attributes/repo.rb
+++ b/attributes/repo.rb
@@ -42,6 +42,7 @@ default['ros_buildfarm']['repo']['container_registry_cache_enabled'] = true
 default['ros_buildfarm']['repo']['pulp_worker_count'] = 2
 default['ros_buildfarm']['repo']['enable_pulp_services'] = true
 default['ros_buildfarm']['rpm_repos']['rhel']['8'] = %w[x86_64]
+default['ros_buildfarm']['rpm_repos']['bootstrap_url'] = 'http://repos.ros.org/repos/$distname/ros_bootstrap/$releasever/$basearch/'
 default['ros_buildfarm']['rpm_upstream_repos']['bootstrap']['rhel']['8'] = Hash[
   architectures: %w[x86_64],
   binary: 'http://repos.ros.org/repos/rhel/ros_bootstrap/8/$basearch/',
--- a/attributes/repo.rb
+++ b/attributes/repo.rb
@@ -42,7 +42,7 @@ default['ros_buildfarm']['repo']['container_registry_cache_enabled'] = true
 default['ros_buildfarm']['repo']['pulp_worker_count'] = 2
 default['ros_buildfarm']['repo']['enable_pulp_services'] = true
 default['ros_buildfarm']['rpm_repos']['rhel']['8'] = %w[x86_64]
-default['ros_buildfarm']['rpm_repos']['bootstrap_url'] = 'http://repos.ros.org/repos/$distname/ros_bootstrap/$releasever/$basearch/'
+default['ros_buildfarm']['rpm_bootstrap_url'] = 'http://repos.ros.org/repos/$distname/ros_bootstrap/$releasever/$basearch/'
 default['ros_buildfarm']['rpm_upstream_repos']['bootstrap']['rhel']['8'] = Hash[
   architectures: %w[x86_64],
   binary: 'http://repos.ros.org/repos/rhel/ros_bootstrap/8/$basearch/',
--- a/attributes/repo.rb
+++ b/attributes/repo.rb
@@ -42,7 +42,7 @@ default['ros_buildfarm']['repo']['container_registry_cache_enabled'] = true
 default['ros_buildfarm']['repo']['pulp_worker_count'] = 2
 default['ros_buildfarm']['repo']['enable_pulp_services'] = true
 default['ros_buildfarm']['rpm_repos']['rhel']['8'] = %w[x86_64]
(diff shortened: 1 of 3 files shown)


recipes/repo.rb
```diff
@@ -284,6 +284,102 @@
...
+  uri 'ppa:osrf/createrepo-agent'
+end
+
+package 'createrepo-agent'
+
+package 'createrepo-c'
+
+package 'rpm'
+
+package 'socat'
```
nuclearsandwich: What introduces the socat and rpm dependencies? Are we using those tools directly or are they dependencies of optional features in createrepo-agent that we're using?

nuclearsandwich: An aside on the formatting, I don't feel strongly about any specific option including leaving it as is, but I don't see a need to add empty lines between one-line resource declarations. It's also possible to use a short loop to express installation for a collection of related packages.

```suggestion
package 'createrepo-agent'
package 'createrepo-c'
package 'rpm'
package 'socat'
```

```suggestion
%w[createrepo-agent createrepo-c rpm socat].each do |pkg|
  package pkg
end
```

recipes/repo.rb
```diff
@@ -284,6 +284,102 @@
...
+
+node['ros_buildfarm']['rpm_repos'].each do |dist, versions|
+  dist_dir = "/var/repos/#{dist}"
+
+  %w(building testing main).each do |repo|
+    repo_dir = "#{dist_dir}/#{repo}"
+    file "#{repo_dir}/RPM-GPG-KEY-ros-#{repo}" do
+      action :delete
+    end
```
nuclearsandwich: What introduced the file that this is deleting? I don't see it on the main production host.

recipes/repo.rb
```diff
@@ -284,6 +284,102 @@
...
+end
+
+if not node['ros_buildfarm']['rpm_bootstrap_url'].empty?
+  file "/home/#{agent_username}/ros_bootstrap_rpm_urls.txt" do
+    owner agent_username
+    group agent_username
+    mode '0644'
+    content node['ros_buildfarm']['rpm_bootstrap_url']
+  end
+end
```
nuclearsandwich: Since the filename is plural should we make the attribute match plurality as list by default?

```suggestion
if not node['ros_buildfarm']['rpm_bootstrap_urls'].empty?
  file "/home/#{agent_username}/ros_bootstrap_rpm_urls.txt" do
    owner agent_username
    group agent_username
    mode '0644'
    content node['ros_buildfarm']['rpm_bootstrap_urls'].join("\n")
  end
end
```

attributes/repo.rb
```diff
@@ -42,6 +42,7 @@
 default['ros_buildfarm']['repo']['pulp_worker_count'] = 2
 default['ros_buildfarm']['repo']['enable_pulp_services'] = true
 default['ros_buildfarm']['rpm_repos']['rhel']['8'] = %w[x86_64]
+default['ros_buildfarm']['rpm_bootstrap_url'] = 'http://repos.ros.org/repos/$distname/ros_bootstrap/$releasever/$basearch/'
```
nuclearsandwich:
```suggestion
default['ros_buildfarm']['rpm_bootstrap_urls'] = ['http://repos.ros.org/repos/$distname/ros_bootstrap/$releasever/$basearch/']
```

Propose making this plural to match the content generated by this attribute.

nuclearsandwich (commented):
Overall this looks good on a first sweep. I have a few questions to follow up before this gets an approval and I owe some developer docs on the strategy for handling the migration of changed and deprecated resources and attributes through major versions of the cookbook as well as some patches to support that here.

recipes/repo.rb
cottsay: This was a copy/paste, really. We originally hosted the public GPG key here but dropped it in favor of downloading it from Github. This was a pretty long time ago and we can probably just drop it.

recipes/repo.rb
cottsay: The `rpm` tool is needed to sign the packages themselves, which we're not quite doing yet, but I'd like to start doing instead of metadata signing.

The `socat` tool isn't strictly needed, but it can be used to essentially open a shell to `createrepo-agent` to debug problems and perform manual modifications.

nuclearsandwich: > `socat` ... can be used to debug problems and perform manual modifications.

Sounds good. I think a comment to that effect would be welcome but hardly mandatory. In the context of security review socat is one of those tools that's sharp enough to cut yourself on so we'd want sound justification for keeping it on a production host (which we have).

recipes/repo.rb
nuclearsandwich: Got it. If we created that file in a tagged release of this cookbook I'd be inclined to keep it in for the next major bump as part of a formal deprecation but if it never made it into a release I don't think that's necessary.

nuclearsandwich (approved):
Comments have all been address. Thanks!

PR #118, opened 2022-12-01
Remove Pulp configuration

cottsay:
We should decide what artifacts we want to specifically remove, and what we'd like to leave behind as an archive.


nuclearsandwich:
> We should decide what artifacts we want to specifically remove

I think our long term goal is to make it so that it's as if pulp never existed on our production hosts. I don't feel the need to save anything now that we've been running on createrepo-agent. I don't think the pulp repositories contain any packages that aren't in the createrepo-agent equivalents either so I am all for scorched earth. The advantage of doing that is that removing pulp will happen in a major version bump and then a major version bump later we can remove all of the pulp cleanup code.

attributes/repo.rb
```diff
@@ -38,9 +38,7 @@
 # ```
 # ['http://repo.test.ros.org:5000']
 # ```
-default['ros_buildfarm']['repo']['container_registry_cache_enabled'] = true
-default['ros_buildfarm']['repo']['pulp_worker_count'] = 2
-default['ros_buildfarm']['repo']['enable_pulp_services'] = true
+default['ros_buildfarm']['repo']['container_registry_cache_enabled'] = false
```
nuclearsandwich: Is disabling the local registry cache intentional? This was added in #28 to mitigate docker hub rate limit restrictions and I don't think it's strictly pulp-related.

recipes/agent.rb
```diff
@@ -98,12 +98,6 @@
   package vcspkg
 end
 
-# Client utilities for pulp content manager
-package "python3-pip"
-%w[pulp-rpm-client==3.9.0 pulpcore-client==3.9.1].each do |pulppkg|
-  execute "python3 -m pip install #{pulppkg}"
-end
-
```
nuclearsandwich: It's probably worth adding a pulp cleanup stanza
```
# TODO remove on next major version bump
%w[pulp-rpm-client==3.9.0 pulpcore-client==3.9.1].each do |pulppkg|
  execute "python3 -m pip uninstall -y #{pulppkg}"
end
package "python3-pip" do
  action :remove
end
```

recipes/repo.rb
```diff
@@ -374,415 +374,63 @@
...
-cookbook_file "#{pulp_data_directory}/initialize.py" do
-  owner 'pulp'
-  source 'pulp/rpm_repo_init.py'
-  mode '0600'
+service 'postgresql' do
+  action [:stop, :disable]
+  only_if { File.exist?('/usr/lib/systemd/system/postgresql.service') }
 end
 
-package 'redis'
```
j-rivero, line 441: Question about removing the package entry: if the redis package was installed in the system, does it make sense to remove it to avoid possible updates that bring a new version back to the system?

j-rivero (commented):
Great clean up, minor comment on how to deal with installed packages. Can not find any major problem.

recipes/repo.rb
cottsay: I was concerned that it might be possible that redis or postgres might be installed for other reasons, and I didn't want to break anything.

If you're confident that they're not otherwise needed, we can remove the packages.

j-rivero: > If you're confident that they're not otherwise needed, we can remove the packages.

Do not know the deployment of the ROS repo good enough to say that nothing else is using redis. Fine to go with the safest approach.

j-rivero (approved):

recipes/repo.rb
nuclearsandwich: > I was concerned that it might be possible that redis or postgres might be installed for other reasons, and I didn't want to break anything.

Is the concern about machines that may be used for other purposes beyond ROS buildfarm repository hosts (i.e. that we'd be removing something which the operator may have wanted installed or started relying on for other tools) or that the buildfarm relies on it for other items?

We aren't using Redis or Postgres for other build farm operations and so I would be in favor of removing these packages in this cleanup. With our intended release model we can do it in two ways:

* publish a new patch release that removes pulp without removing dropped packages, then remove them in a new minor release listing the removal as a breaking change, the removal can then be dropped in a future minor, completing the pulp cleanup arc.
* publish a new patch release that removes these packages as internal to pulp and do so with a warning that if if you're using redis or postgres for other services on the system you'll need to set an attribute to block their removal. The removal can then be dropped in a future minor, completing the pulp cleanup arc.
* Leave the packages "unmanaged" and recommend manual removal steps for dedicated systems.

nuclearsandwich (commented):
Ooof. I have 3 pending review comments on this PR that were never submitted. I can find one of them in the current diff. I'll mark the other two resolved once I see them.

recipes/agent.rb
nuclearsandwich: Done in https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/118/commits/8e767cb5f363f1aa463353e16ad9616db106ea79

nuclearsandwich:
Superseded by #135.

PR #120, opened 2022-12-02
Use a test public key with no expiration date.

nuclearsandwich:
The repo host CI is failing during reprepro setup because the GPG key used for tests hit its two year anniversary. This updates the public key so it no longer has an expiration date.


nuclearsandwich:
Jenkins CI will likely fail due to #119.

nuclearsandwich:
Superseded by #131.

PR #128, opened 2023-03-16
Only delete S.gpg-agent if it is a symlink

cottsay:
This deletion reverts the use of the gpg valult for jenkins-agent that was previously implemented. If there is already a gpg-agent process running on the host when this runs, it tries and fails to remove the socket there even though it's not a symlink. If it's not a symlink, then just leave it alone.

My previous workaround for this issue was to stop the `gpg-agent` process for the `jenkins-agent` user prior to deploying.


nuclearsandwich (approved):
Nice catch on the cleanup!

PR #131, opened 2023-06-20
Push out expiration of test keys.

nuclearsandwich:
I'm not sure if it makes sense to make these indefinite but for the time being I've just bumped it three years. I have no justification for three versus one or two or five.


cottsay (commented):
Post hoc :heavy_check_mark:

PR #133, opened 2024-03-21
Use cinc for test kitchen.

nuclearsandwich:
Avoids using encumbered chef binaries which currently fail pending a license acceptance when running locally.


nuclearsandwich:
The last failure here should be resolved by #119. I'll get that in first and rebase this.

claraberendsen (approved):
LGTM!

PR #135, opened 2024-07-18
Drop pulp support and remove pulp support configuration and dependencies.

nuclearsandwich:
This is a rebase of #118 with some modifications:

* Address https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/118#discussion_r1579698429 by removing the redis and postgresql packages (but does not "purge" them, removing configuration and data as well although I am tempted to do so).

* Address https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/118#discussion_r1038624229 and the test failure by re-enabling the docker-registry by default.

* Document that the pulp cleanup stanzas can be removed in the next major release.


recipes/repo.rb
```diff
@@ -374,415 +374,70 @@
...
+file '/usr/local/bin/systemd-docker' do
+  action :delete
 end
 
-package 'redis'
-service 'redis' do
-  action [:start, :enable]
+service 'postgresql' do
+  action [:stop, :disable]
+  only_if { File.exist?('/usr/lib/systemd/system/postgresql.service') }
```
j-rivero: Double checking: I can not find this path on my focal system. Was it produced by a custom previous installation?, should it `/lib/systemd/system/postgresql.service`?

Same for redis below.

recipes/agent.rb
```diff
@@ -122,9 +122,10 @@
 end
 
 # Client utilities for pulp content manager
-package "python3-pip"
-%w[pulp-rpm-client==3.9.0 pulpcore-client==3.9.1].each do |pulppkg|
-  execute "python3 -m pip install #{pulppkg}"
+%w[pulp-rpm-client pulpcore-client].each do |pulppkg|
```
j-rivero, line 126: Maybe worth marking this to be removed in the next release.

j-rivero (approved):
Can not detect any big problem, just a check on a system path in a comment I left before.

recipes/repo.rb
nuclearsandwich: > Double checking: I can not find this path on my focal system. Was it produced by a custom previous installation?, should it /lib/systemd/system/postgresql.service?

In my test systems `/usr/lib` and `/lib` are one and the same. But the reported value in `systemctl status postgresql.service` canonicalizes "bare" `/lib` so I'll do likewise.

recipes/agent.rb
nuclearsandwich: Done in 282425188618b1619de5d85871985f072e76abf8

recipes/repo.rb
nuclearsandwich: Done in 282425188618b1619de5d85871985f072e76abf8

nuclearsandwich:
In addition to the CI checks, I also confirmed that this branch can be converged on top of `0.5.1` validating it for inclusion in a `0.6.0` release.

This also converges cleanly on `0.5.0` which means that we could technically release it in a `0.5.2` although since the scope of this change is to remove an entire feature (that was switched off by default in `0.5.1` I think a tick-tock approach across two major versions is warranted.

PR #137, opened 2024-07-20
Add handler script to support signing apt repository with multiple keys.

nuclearsandwich:
Reprepro will handle signing if an individual key is specified or if the default gpg can be used, but signing a release with multiple keys requires the use of a handler script.

The handler script is added as a static script which relies on a new shell configuration file to set variables dictating the keys used via adding them to KEY_OPTS.

Additional GPG options can also be set in the GPG_OPTS variable although this is currently unused.

The config file is intended to be used as a chef template which will incorporate the keys to be used.

### Caveats

I actually wrote this on an omnibus branch that included #136  and #135 because the pulp removal simplifies a lot of the GPG configuration. I'm not sure if this branch rebased onto latest will build.

Unsupported as of now is the ability to use multiple signatures on the rpm repositories.
For the time being they are only signed with the key designated "default".

It would be nice to try and resurrect the gpg-vault setup from #55, reverted in #82 because of consistency issues. I don't want to block on that but I would like to port the GPG vault approach forward and see if we have the same issues on a test farm in a future PR.


test/integration/repo/repo_test.rb
```diff
@@ -21,3 +21,16 @@
...
+describe file('/var/repos/repos.key') do
+  it { should exist }
+  it { should be_symlink }
+end
+
+describe file('/var/repos/ubuntu/main/dists/focal/InRelease') do
+  it { should exist }
+end
+
+describe command('su - jenkins-agent -c "gpg --verify /var/repos/ubuntu/main/dists/focal/InRelease"') do
```
nuclearsandwich, line 34: Running this as the jenkins-agent user, whose keyring has the public and private signing keys on it, verifies that the signatures are all valid.

test/integration/data_bags/ros_buildfarm_repository_signing_keys/test-beta.json
```diff
@@ -0,0 +1,7 @@
+{
+  "id": "test-beta",
+  "default": true,
```
nuclearsandwich, line 3: I stuck the default tag on the "middle" key to make sure that it works when not the first or the last. The "default" key is the one that our ros-archive-keyring package should set `SignedBy=` for.

templates/reprepro-sign-config.sh.erb
```diff
@@ -0,0 +1,10 @@
+# Configuration variables for reprepro-sign.sh
+GPG_OPTS=""
+
+KEY_OPTS=""
+<% if @signing_key_ids.size != 0 -%>
+<% @signing_key_ids.each do |fingerprint| -%>
+KEY_OPTS="$KEY_OPTS -u <%= fingerprint %>"
+<% end -%>
+<% end -%>
```
nuclearsandwich, line 9: This could be templated out another way, but I think this is a reasonable pattern.

I realize as I write that the test for a non-zero size is vestigial from when I was planning to make this
```suggestion
<% if @signing_key_ids.size != 0 -%>
KEY_OPTS="$KEY_OPTS -u <%= @signing_key_ids.join(" -u ") %>"
<% end -%>
```

and thus can be removed. Of course, if this version is preferred then the loop can be removed instead.

recipes/repo.rb
```diff
@@ -173,12 +238,15 @@
   only_if { ::File.symlink?("/home/#{agent_username}/.gnupg/S.gpg-agent") }
 end
 
-# Set up ssh authorized keys for publish over ssh.
-directory "/home/#{agent_username}/.ssh" do
```
nuclearsandwich, line 177: This is diff artifacting. There is no relationship to the added cookbook_file. This section was moved up to line 115 to allow the ssh directory to be used for GPG private keys.

recipes/repo.rb
```diff
@@ -111,35 +111,100 @@
   triggers_reload true
```
nuclearsandwich, line 111: This file appears to suffer from a pretty clumsy rebase. Please hold off on review of it for now.

recipes/repo.rb
```diff
@@ -173,12 +238,15 @@
   only_if { ::File.symlink?("/home/#{agent_username}/.gnupg/S.gpg-agent") }
 end
 
-# Set up ssh authorized keys for publish over ssh.
-directory "/home/#{agent_username}/.ssh" do
+cookbook_file "/home/#{agent_username}/reprepro-sign.sh" do
```
j-rivero, line 241: The path is used also in a template command a bit below. Could be a good idea to store it in a single place/var to facilitate maintenance and changes.

recipes/repo.rb
```diff
@@ -268,12 +336,22 @@
   mode '0600'
   variables Hash[
     architectures: node['ros_buildfarm']['apt_repos']['architectures'],
-    signing_key: gpg_key['fingerprint'],
+    signing_key: "! /home/#{agent_username}/reprepro-sign.sh",
```
j-rivero, line 339:
```suggestion
    # The ! + script notation is defined in the reprepro manpage when using external scripts 
    signing_key: "! /home/#{agent_username}/reprepro-sign.sh",
```

recipes/repo.rb
```diff
@@ -111,35 +111,100 @@
...
-  append true
-  members [agent_username]
-  action [:manage]
+
+gpg_keys = search('ros_buildfarm_repository_signing_keys', "id:#{node.chef_environment}-*")
+if gpg_keys.size == 0
+  Chef::Log.warn("Support for multiple GPG keys has changed the layout if this data bag.")
+  Chef::Log.warn("Please see the Changelog entry Forthcoming for details.")
+  Chef::Log.warn("The sole key will be assumed the default.")
+  gpg_keys = search('ros_buildfarm_repository_signing_keys', "id:#{node.chef_environment}")
```
j-rivero, line 126: same command than in 121, is this intentional?

test/integration/repo/repo_test.rb
j-rivero: Could be the case that this verifies that only a single signature is present and valid but we are not verifying in the PR that we are indeed signing the file with multiple keys?

j-rivero (commented):
Left some comments and suggestions. Mostly fine, great work!

There are valid problems reported in the CI but you probably assumed that being this in Draft status.

PR #142, opened 2025-01-13
24.04 migration: Authentication Strategy

Crola1702:
# Description

This improves the management of Jenkins authentication by removing any reference to old jenkins cookbook resources. Also, uses groovy scripts put in init.groovy.d directory for initializing authentication and configuring authorization

### Changes 
* Created `/var/lib/jenkins/init.groovy.d` to store `auth_strategy.groovy` defined by the selected authorization strategy defined in the cookbook attributes
* Replaced the use of `jenkins_script` and `jenkins_user` to use groovy scripts for the auth_strategy

PR #143, opened 2025-01-13
24.04 migration: Credentials

Crola1702:
# Description

This PR replaces the use of `jenkins_*_credentials` resources with Groovy scripts to handle the use of credentials more dynamically and flexibly.

Also, it creates `credentials_config.groovy` inside `/var/lib/jenkins/init.groovy.d/` (see #142)

### DIFF
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -238,7 +238,6 @@ elsif node.default['ros_buildfarm']['jenkins']['auth_strategy'] == 'default'
   matrix_auth_permissions_script = <<~GROOVY
       import hudson.security.ProjectMatrixAuthorizationStrategy
 
-      def jenkins = Jenkins.getInstance()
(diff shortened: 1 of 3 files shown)


recipes/jenkins.rb
```diff
@@ -342,32 +342,132 @@ def jenkins = Jenkins.getInstance()
...
+        )
+      } else {
+        credentials_store.addCredentials(global_domain, credentials)
+      }
+    GROOVY
+end
+
+file '/var/lib/jenkins/init.groovy.d/credentials_config.groovy' do
+  content credentials_scripts.join("\n")
+  mode '0755'
```
claraberendsen: I think we want a `0500` permission here to allow the jenkins user to read and execute, but not write. 
It would prevent other users from reading the content of the groovy script.
Do we know if Jenkins needs to read it as well or if just executes what's under init.groovy?

PR #147, opened 2025-03-03
Cookbook 24.04 migration

Crola1702:
# Description

This PR includes:
* Multiple updates to the Jenkins configuration recipe and plugins management for 24.04 migraion
* Improves the management of Jenkins authentication by removing any reference to old jenkins cookbook resources. Also, uses groovy scripts put in init.groovy.d directory for initializing authentication and configuring authorization
* Replaces the use of `jenkins_*_credentials` resources with Groovy scripts to handle the use of credentials more dynamically and flexibly.
* Updates the agents to work with Ubuntu 24.04 by:

## Changes:
* Updated Jenkins version to `2.479.2` 
* Added a new attribute for JDK version 
* Changed the included recipe from `jenkins::master` to `jenkins::jenkins`
* Updated numerous Jenkins plugins to their latest versions to ensure compatibility and security
* Improved the removal of obsolete plugins by using a filtered grep command to delete specific `.jpi` files.
* Added a new recipe to manage plugin installation using a `plugins.txt` template and the Jenkins plugin manager tool.
* Updated Jenkins configuration in `jenkins.yaml.erb` to reflect changes in plugin names and configurations.
* Created `/var/lib/jenkins/init.groovy.d` to store `auth_strategy.groovy` defined by the selected authorization strategy defined in the cookbook attributes
* Replaced the use of `jenkins_script` and `jenkins_user` to use groovy scripts for the auth_strategy
* Also, it creates `credentials_config.groovy` inside `/var/lib/jenkins/init.groovy.d/` (see #142)
* Configuring jdk version to be 21 for agents.
* Add bzip2 package in agents (not present by default in 24.04)


attributes/jenkins.rb
```diff
@@ -23,4 +23,7 @@
 # Last version supporting Java 8
 #default['jenkins']['master']['version'] = '2.346.1'
 # Last version supporting sysvinit scripts
-default['jenkins']['master']['version'] = '2.319.3'
+default['jenkins']['master']['version'] = '2.492.2'
```
nuclearsandwich, line 26: We'll eventually want to drop this pin entirely to allow unattended-upgrades to update Jenkins.

recipes/agent.rb
```diff
@@ -111,6 +112,9 @@
 
 package 'python3-empy'
 
+# Install bzip2 as it's not present in ubuntu noble by default
```
nuclearsandwich: Can you note what this dependency serves?

recipes/jenkins.rb
```diff
@@ -455,13 +455,138 @@ def jenkins = Jenkins.getInstance()
...
-file '/var/lib/jenkins/init.groovy.d/credentials_config.groovy' do
+file '/var/lib/jenkins/init.groovy.d/02-credentials_config.groovy' do
   content credentials_scripts.join("\n")
   mode '0500'
   owner 'jenkins'
   group 'jenkins'
 end
 
+file '/var/lib/jenkins/init.groovy.d/99-approved_signatures.groovy' do
+  content <<~GROOVY
```
nuclearsandwich: This is a long enough set of contents that I'd consider making it a cookbook file (or even a template) rather than embedding the file contents here.

nuclearsandwich (approved):

claraberendsen (approved):

recipes/jenkins.rb
Crola1702: Done in 950603d

recipes/agent.rb
Crola1702: Done. It is used by https://github.com/ros-infrastructure/ros_buildfarm/blob/896e5e6cc90a0eb4fbfc2e31c22661e6adb08230/ros_buildfarm/templates/ci/ci_job.xml.em#L167 (added in a5158da)

attributes/jenkins.rb
Crola1702: I opened an issue to track this into a different PR: See #158

PR #151, opened 2025-03-11
Increase timeout for jenkins service unit

claraberendsen:
### Description 

This PR increases the timeout of the systemd unit so all init.groovy scripts can run on a jenkins start.


Crola1702 (commented):
I think we're not pointing to the right service. The timeout is occurring on `jenkins.service` instead.

Maybe we should instead create a template for `jenkins.service.d/override.conf` and use `TimeoutStartSec`?

```
# Configures the time to wait for start-up. If Jenkins does not signal start-up
# completion within the configured time, the service will be considered failed
# and will be shut down again. Takes a unit-less value in seconds, or a time span
# value such as "5min 20s". Pass "infinity" to disable the timeout logic.
#TimeoutStartSec=90
```

Crola1702:
I'm closing this PR in favor of #157

PR #152, opened 2025-03-11
Add groovy script to start in shutdown mode.

nuclearsandwich:
We've used this on each build farm for a while now but it was never part of the chef config.

This is extremely helpful in preventing jobs from running during multiple restarts for deployment updates.

### DIFF
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -117,7 +117,7 @@ directory '/var/lib/jenkins/init.groovy.d' do
   group 'jenkins'
 end
 
-file '/var/lib/jenkins/init.groovy.d/start-quiet.groovy' do
+file '/var/lib/jenkins/init.groovy.d/03-start-quiet.groovy' do
   content <<-GROOVY
 import jenkins.model.*
 
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -112,6 +112,24 @@ node['ros_buildfarm']['jenkins']['plugins'].each do |plugin, ver|
   end
 end
 
+directory '/var/lib/jenkins/init.groovy.d' do
+  owner 'jenkins'
+  group 'jenkins'
+end
+
+file '/var/lib/jenkins/init.groovy.d/start-quiet.groovy' do
+  content <<-GROOVY
+import jenkins.model.*
+
+Jenkins.get().doQuietDown()
+  GROOVY
+
+  mode '0660'
+  owner 'jenkins'
+  group 'jenkins'
+end
+
+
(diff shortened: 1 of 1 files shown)


recipes/jenkins.rb
```diff
@@ -112,6 +112,24 @@
   end
 end
 
+directory '/var/lib/jenkins/init.groovy.d' do
+  owner 'jenkins'
+  group 'jenkins'
+end
+
+file '/var/lib/jenkins/init.groovy.d/start-quiet.groovy' do
```
Crola1702: I proposed using number prefixes to prevent scripts dependency errors:

See https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/147.

```suggestion
file '/var/lib/jenkins/init.groovy.d/03-start-quiet.groovy' do
```

recipes/jenkins.rb
```diff
@@ -112,6 +112,24 @@
...
+  content <<-GROOVY
+import jenkins.model.*
+
+Jenkins.get().doQuietDown()
+  GROOVY
+
+  mode '0660'
```
Crola1702: Also, we use mode 500: https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/147/files#diff-e2bbe1895ea17f8e385d377b5c5b06b4d5720bbce22252f1880269a305304a1aR460

```suggestion
  mode '0500'
```

Crola1702 (changes requested):
Just some comments that propose to align the `init.groovy.d/*` conventions used in https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/147/

recipes/jenkins.rb
```diff
@@ -112,6 +112,24 @@
...
+directory '/var/lib/jenkins/init.groovy.d' do
+  owner 'jenkins'
+  group 'jenkins'
+end
+
+file '/var/lib/jenkins/init.groovy.d/start-quiet.groovy' do
+  content <<-GROOVY
+import jenkins.model.*
+
+Jenkins.get().doQuietDown()
```
Crola1702, line 124: Also, do you think it would be worth to add this as a cookbook_file below jenkins/groovy-scripts with the rest of scripts present in init.groovy.d?

claraberendsen: I think this would be a good idea, I don't think it would reduce code length but it means all init.groovy scripts would be in the same place.

claraberendsen (approved):
Left a nit comment for improvement. Other than that it looks good! Thanks for pushing this @Crola1702

PR #157, opened 2025-03-20
Add a jenkins service override with TimeoutStartSec

Crola1702:
When jenkins.service starts for the first time, it needs to execute all the scripts below `init.groovy.d`. The sum of all these scripts take a lot of time (specially `99-approved_signatures.groovy`).

The default timeout for the unit is 90s, but the execution takes more time than that (at least 2 minutes)

This PR adds a template for `jenkins.service.d/override.conf` which contains `TimeoutStartSec` defined as 180s (3 minutes)

Also, I removed a restart that was causing a `systemd unit start limit hit` error when restarting jenkins

### DIFF
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -104,7 +104,7 @@ template '/etc/systemd/system/jenkins.service.d/override.conf' do
   owner node['jenkins']['master']['user']
   group node['jenkins']['master']['group']
   variables Hash[
-    timeout_start_sec: 360
+    timeout_start_sec: 180
   ]
 end
 
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -92,6 +92,22 @@ end
 
 include_recipe 'jenkins::jenkins'
 
+# Increase timeout of jenkins systemd unit
+directory '/etc/systemd/system/jenkins.service.d' do
+  mode '0755'
+  owner 'root'
+  group 'root'
+end
+
+template '/etc/systemd/system/jenkins.service.d/override.conf' do
+  source 'jenkins/jenkins-service-override.conf.erb'
+  owner node['jenkins']['master']['user']
+  group node['jenkins']['master']['group']
+  variables Hash[
+    timeout_start_sec: 360
+  ]
+end
+
 # Set up authentication
 chef_user = search('ros_buildfarm_jenkins_users', 'chef_user:true').first
 node.run_state[:jenkins_username] = chef_user['username']
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -99,13 +99,14 @@ directory '/etc/systemd/system/jenkins.service.d' do
   group 'root'
 end
 
-template '/etc/systemd/system/jenkins.service.d/override.conf' do
-  source 'jenkins/jenkins-service-override.conf.erb'
-  owner node['jenkins']['master']['user']
-  group node['jenkins']['master']['group']
-  variables Hash[
-    timeout_start_sec: 180
-  ]
+cookbook_file '/etc/systemd/system/jenkins.service.d/500-timeout.conf' do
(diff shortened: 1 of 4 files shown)


recipes/jenkins.rb
```diff
@@ -92,6 +92,22 @@
...
+directory '/etc/systemd/system/jenkins.service.d' do
+  mode '0755'
+  owner 'root'
+  group 'root'
+end
+
+template '/etc/systemd/system/jenkins.service.d/override.conf' do
+  source 'jenkins/jenkins-service-override.conf.erb'
+  owner node['jenkins']['master']['user']
+  group node['jenkins']['master']['group']
```
claraberendsen: Should this be owned by root as the folder as well? Or is there a reason for this being owned by the jenkins user?

claraberendsen (changes requested):

recipes/jenkins.rb
```diff
@@ -92,6 +92,22 @@
...
 include_recipe 'jenkins::jenkins'
 
+# Increase timeout of jenkins systemd unit
+directory '/etc/systemd/system/jenkins.service.d' do
+  mode '0755'
+  owner 'root'
+  group 'root'
+end
+
+template '/etc/systemd/system/jenkins.service.d/override.conf' do
```
nuclearsandwich: This will collide with the file created in the jenkins cookbook: https://github.com/ros-infrastructure/cookbook-jenkins/blob/9435bd9121d0c38b745a5548b4f1ea8365c7f43e/recipes/jenkins.rb#L67-L73

systemd can actually process multiple override files. Any *.conf file in the directory. So we could create a single `500-timeout-start.conf` with fixed content and avoid using a template file here entirely.

recipes/jenkins.rb
nuclearsandwich: Yes, it is read by systemd and not the jenkins user.

recipes/jenkins.rb
```diff
@@ -92,6 +92,22 @@
...
+  source 'jenkins/jenkins-service-override.conf.erb'
+  owner node['jenkins']['master']['user']
+  group node['jenkins']['master']['group']
+  variables Hash[
+    timeout_start_sec: 180
+  ]
+end
```
nuclearsandwich, line 111: In order to take effect this needs to trigger a systemd daemon-reload: https://github.com/ros-infrastructure/cookbook-jenkins/blob/9435bd9121d0c38b745a5548b4f1ea8365c7f43e/recipes/jenkins.rb#L72-L73

recipes/jenkins.rb
```diff
@@ -207,12 +223,6 @@ def jenkins = Jenkins.getInstance()
     group 'jenkins'
   end
 
-  # Restart jenkins after updating the security realm otherwise running without
-  # authentication yields 403 errors when configuring.
-  service 'jenkins' do
-    action :restart
-  end
-
   # Aggregate permissions to assign to each user with a groovy script.
```
nuclearsandwich: This seems like treatment of a symptom rather than the underlying problem.

Changing one or both of TimeoutStopSec or RestartSec seems more appropriate since normal cookbook convergence shouldn't cause failures starting during configuration.

nuclearsandwich (changes requested):

recipes/jenkins.rb
Crola1702: I used RestartSec on the coookbook_file

recipes/jenkins.rb
```diff
@@ -92,6 +92,23 @@
 
 include_recipe 'jenkins::jenkins'
 
+# Increase timeout of jenkins systemd unit
```
claraberendsen, line 95: (nit) It would be good to add rationale behind the increase of the timeout in this comment to have a pointer as to why this was added.

claraberendsen (approved):

PR #159, opened 2025-04-17
Restart ssh after updating host keys.

nuclearsandwich:
I did this completely blind and haven't yet tested it at all.


Crola1702 (approved):

Crola1702:
Just added c587ffd. `notifies` was outside a resource block.

I tested it and this change shows at the end of the chef run:
```
  * service[ssh] action restart
    - restart service service[ssh]
```

PR #160, opened 2025-04-17
Run base GitHub Action on 24.04.

nuclearsandwich:


Crola1702 (approved):

nuclearsandwich:
@Crola1702 I'm going to merge this as-is but I think the kitchen CI has found a gap in our configuration when not using AWS images as the base.

PR #169, opened 2025-12-10
Add openjdk- as unatteded upgrades blacklist

Crola1702:
This is a mitigation to prevent https://github.com/osrf/buildfarm-tools/issues/189#issuecomment-3596997307 from happening again

### DIFF
--- a/attributes/agent.rb
+++ b/attributes/agent.rb
@@ -34,4 +34,4 @@ default['ros_buildfarm']['agent']['labels'] = %w(buildagent)
 default['ros_buildfarm']['agent']['jdk_version'] = 21
 
 # Prevents docker and containerd from getting updates and restarting mid build. See https://github.com/ros2/ci/issues/702
-default['ros_buildfarm']['unattended_upgrades']['package_blacklist'] = %w[docker.io containerd]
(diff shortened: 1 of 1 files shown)


claraberendsen:
@Crola1702 the failures seem unrelated to this change. But do you know what is causing them?

nuclearsandwich-ai (approved):

attributes/agent.rb
```diff
@@ -34,4 +34,4 @@
 default['ros_buildfarm']['agent']['jdk_version'] = 21
 
 # Prevents docker and containerd from getting updates and restarting mid build. See https://github.com/ros2/ci/issues/702
-default['ros_buildfarm']['unattended_upgrades']['package_blacklist'] = %w[docker.io containerd]
+default['ros_buildfarm']['unattended_upgrades']['package_blacklist'] = %w[docker.io containerd openjdk-]
```
nuclearsandwich, line 37: I assume that the pattern match means that this would block all packages matching the glob `openjdk-*` but it may be worth a comment to explain the pattern syntax here out-of-context.

nuclearsandwich (approved):

PR #177, opened 2026-06-15
Make CI green 

claraberendsen:
### Description 
 
 This PR modernizes the cookbook-ros-buildfarm to support Ubuntu 24.04 LTS, the latest Jenkins and Chef/Cinc versions, and fixes CI failures caused by version and configuration drift.

Notable changes: 

- Use new jenkins cookbook instead of upstream chef: https://github.com/ros-infrastructure/cookbook-jenkins/
- Bump cinc version to latest (19) and point to Noble as base for CI. 
- Address configuration drift changes on the CasC plugin and swarm agents. 
- Enforce idempotency checks on CI by default. 
- Repository signing key `expire` for test data_bags is set to never to prevent failure of CI when a key expires. The expiry date for the key was 19th of June of 2026.

### Changes 
<details>

1f9e4b3 — bump to 24.04 and latest chef stable: update CI and test kitchen to Ubuntu 24.04, bump Chef version to 19
f398255 — add reference to new jenkins cookbook: switch Berksfile dependency from pinned Supermarket version to ros-infrastructure/cookbook-jenkins git latest branch
41e4297 — use cinc bin instead of chef: change test kitchen binary path to /opt/cinc/bin/cinc-solo
661ae27 — modify needrestart config to use the directory structure (agent): replace inline needrestart.conf modification with a drop-in file under /etc/needrestart/conf.d/jenkins-agent.conf, per Ubuntu 24.04 needrestart changes
07f2aea — use directory structure to configure needrestart (jenkins): same drop-in approach for /etc/needrestart/conf.d/jenkins.conf; also changes restart notifications from :immediately to :delayed to avoid multiple Jenkins restarts in a single converge
3265d67 — fix casc config to match latest jenkins: simplify JCasC crumbIssuer from verbose nested form to crumbIssuer: 'standard'; remove myViewsTabBar (now default)
9be3bd9 — update parameters to match swarm > 3.x: rename -master to -url in jenkins-agent.service.erb
6eb24f9 — add idempotency test and multiple converge to ci: add idempotency_test: true and multiple_converge: 2 to test kitchen config
fc257fc — update Jenkins to latest: bump Jenkins to 2.555.3 to prevent the "prevent downgrade" guardrail from failing on repeated converges
1f6e429 — use https instead of ssh for git operations
d5b4a7c — fix agent test: update integration test to check for openjdk-21-jdk-headless instead of openjdk-8-jdk-headless
a208221 — remove duplicate entry for agent recipe: drop redundant recipe[ros_buildfarm::agent] from the jenkins test suite (already included via the jenkins recipe run list)
86a43e3 — delay jenkins restarts: change two notifies :restart calls in recipes/jenkins.rb from :immediately to :delayed so a single converge only restarts Jenkins once
e7c00ae - update test GPG key to never expire to prevent maintenance burden for key expiration on CI. 

</details>


recipes/agent.rb
```diff
@@ -154,8 +155,8 @@
     user_home: agent_homedir,
     labels: node['ros_buildfarm']['agent']['labels'],
   ]
-  notifies :restart, 'service[jenkins-agent]'
```
j-rivero, line 157: Any particular reason to remove the restart on /etc/default/jenkins-agent changes?

recipes/jenkins.rb
```diff
@@ -551,9 +557,7 @@ def jenkins = Jenkins.getInstance()
   append true
   action :manage
 end
-service 'jenkins' do
-  action :start
```
j-rivero, line 511: Jenkins is topped in https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/177/changes#diff-e2bbe1895ea17f8e385d377b5c5b06b4d5720bbce22252f1880269a305304a1aL534 and unless other configure restart it it might end up stopped?

recipes/agent.rb
claraberendsen: Addressed in [9d8377e](https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/177/commits/9d8377e7de03a16e570174afc7aadb3fedaf9b32)

recipes/jenkins.rb
claraberendsen: The Jenkins instance is restarted here https://github.com/ros-infrastructure/cookbook-ros-buildfarm/blob/9d8377e7de03a16e570174afc7aadb3fedaf9b32/recipes/jenkins.rb#L251-L255. And there is a smoke test in CI that verifies that Jenkins is running https://github.com/ros-infrastructure/cookbook-ros-buildfarm/blob/86a43e3cacb608b04feb59cddeda477a881c50a6/test/integration/jenkins/jenkins_test.rb#L1-L3 so I wouldn't expect this to fail with this removal. 

I removed it here because CI was failing to converge due to systemd failed with  Jenkins restarting too often. By `:delayed` all restarts we ensure these are run only once at the end of the converge phase.

j-rivero: This is an interesting case, note that the stop is produced in line 543 so unless the delayed cases do it, the restart on 254 should not produce the start.

There are two delayed clauses in this file, I think that are attached to a template generation. If that template generation does not change, a second converge should not triggered the restart.

Now, what is triggering the restart? I think that the plugin recipe is the one pulling the restart everytime since chef probably consider it as "updated" on every single run. The `execute` has no idem-potency clause "only_if / not_if".

Maybe the best option could be move that plugin command to run only on plugin list changes via template, drafting:

```ruby
 template '/var/lib/jenkins/plugins/plugins.txt' do
    source 'plugins.txt.erb'
    notifies :run, 'execute[install-jenkins-plugins]', :immediately
  end

  execute 'install-jenkins-plugins' do
    command %w[
      java -jar /usr/local/jenkins/jars/jenkins-plugin-manager.jar
     ... lalala
    ]
    action :nothing  # ← not run unless notified
    notifies :restart, 'service[jenkins]', :delayed
  end
```

This should give us idempotency on the whole resource. This probably is out of the scope of the PR #359.

Back to our original bussiness: the whole thing works because of a side-effect under my eyes, we can live with it here and create a PR that address the whole thing with the patch in this comment + restore the "restart" or find a better solution.

kitchen.dokken.yml
```diff
@@ -11,15 +11,17 @@ transport:
 
 provisioner:
   name: dokken
-  chef_binary: /opt/chef/bin/chef-solo
+  chef_binary: /opt/cinc/bin/cinc-solo
   chef_options: "-Etest"
   chef_log_level: debug
   chef_output_format: doc
+  idempotency_test: true
```
j-rivero: This hits me since I'm pretty sure that the code is not idempotent compliance in some cases. I think that the `idempotency_test` attribute does not exists in the provisioner and is doing nothing. Maybe `enforce_idempotency`? https://github.com/test-kitchen/test-kitchen/pull/875/changes

claraberendsen: This is a good catch. It is indeed enforece_idempotency the correct one. Updated on [9602525](https://github.com/ros-infrastructure/cookbook-ros-buildfarm/pull/177/commits/9602525f34489e4382974f7325e7078f89f564f3).

claraberendsen:
@j-rivero you make two good points. I agree that given that the idempotency check was not working correctly my assumption of the :delayed trigger being sufficient is no longer valid.
While I believe idempotency of the recipes to be important for the confidence in our chef runs, I think getting the recipes to state is a heavier lift than the original goal here. Just looking for example at the output of the Jenkins recipe diff there is a lot to change. 

```
    First cinc run should have reached a converged state.
       Resources updated in a second cinc-client run:
       - apt_package[jenkins]
       - ruby_block[prevent jenkins downgrade]
       - execute[systemctl-daemon-reload]
       - execute[install-jenkins-plugins]
       - execute[ls /var/lib/jenkins/plugins | grep -E "analysis-core.jpi|config-file-provider.jpi|cvs.jpi|icon-shim.jpi" | xargs -r rm]
       - file[/var/lib/jenkins/init.groovy.d/01-auth_strategy.groovy]
       - service[jenkins]
       - file[/var/lib/jenkins/init.groovy.d/01-auth_strategy.groovy]
       - service[jenkins]
       - directory[/var/lib/jenkins/fingerprints]
       - directory[/var/lib/jenkins/fingerprints]
       - group[jenkins-agent]
       - service[jenkins]

```

I'll remove the idempotency requirement from this PR and track it in an issue with a draft PR and make progress on that front. I think having some form of green CI is better than no working CI for this repo.

j-rivero:
> I'll remove the idempotency requirement from this PR and track it in an issue with a draft PR and make progress on that front. I think having some form of green CI is better than no working CI for this repo.

Yep, larger goal than the PR scope, I agree.

PR #182, opened 2026-07-23
Move CI to ubuntu 24.04 and chef 19

Crola1702:
# Description

This PR updates our test infrastructure to match the Ubuntu 24.04 platform and test and Chef 19, it also fixes a string interpolation issue in the Jenkins Groovy script, and optimizes service restart behavior during chef runs

## Changes
- Updated Test Kitchen configuration and CI Workflow from Ubuntu 20.04 to Ubuntu 24.04. Added Chef 19 to the matrix suite
- Changed service restart notifications for configuration file updates from `:immediately` to `:delayed`. This prevents multiple abrupt service restarts during a single run, allowing Chef to converge before restarting Jenkins. (Part of #177)
- Fixed multiline SSH public key interpolation in the Groovy user setup and correct double-quote escaping for "`\n`"

> [!NOTE]
>
> Note that we won't run jenkins CI on 20.04 from now on

### DIFF
--- a/recipes/jenkins.rb
+++ b/recipes/jenkins.rb
@@ -234,7 +234,7 @@ elsif node.default['ros_buildfarm']['jenkins']['auth_strategy'] == 'default'
       }
       password = hudson.security.HudsonPrivateSecurityRealm.Details.fromPlainPassword("#{user['password']}")
       user.addProperty(password)
-      keys = new org.jenkinsci.main.modules.cli.auth.ssh.UserPropertyImpl(#{user['public_keys'].join('\n')})
+      keys = new org.jenkinsci.main.modules.cli.auth.ssh.UserPropertyImpl("""#{user['public_keys'].join("\n")}""")
       user.addProperty(keys)
       user.save()
     GROOVY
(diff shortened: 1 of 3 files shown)


recipes/jenkins.rb
```diff
@@ -234,7 +234,7 @@ def jenkins = Jenkins.getInstance()
       }
       password = hudson.security.HudsonPrivateSecurityRealm.Details.fromPlainPassword("#{user['password']}")
       user.addProperty(password)
-      keys = new org.jenkinsci.main.modules.cli.auth.ssh.UserPropertyImpl(#{user['public_keys'].join('\n')})
+      keys = new org.jenkinsci.main.modules.cli.auth.ssh.UserPropertyImpl("""#{user['public_keys'].join("\n")}""")
```
claraberendsen, line 237: What prompted this change and how was it working correctly before?

Crola1702: > What prompted this change and how was it working correctly before?

A local failure. It Failed because the user public keys were not quoted. 

It was working before because on production we don't use public keys with this authentication method. See for reference: https://github.com/osrf/chef-osrf/tree/latest/data_bags/ros_buildfarm_jenkins_users

claraberendsen (approved):

PR #183, opened 2026-07-24
Update jenkins version to 2.555.3

Crola1702:
# Description

Migrates the Jenkins controller from 2.492.2 to 2.555.3, crossing five LTS baselines (2.504, 2.516, 2.528, 2.541, 2.555). This  a full plugin re-pin, and  (JCasC) fields changes.

## Changes
- Bump jenkins version to `2.555.3`
- Remove `crumbIssuer.standard.excludeClientIPFromCrumb` from `jenkins.yaml.erb`, deprecated as of 2.555.1; JCasC hard-fails startup if it's present.
- Remove `myViewsTabBar` / `viewsTabBar` from the same template, deprecated as of 2.516.x for the same reason.
- Re-pin the full `node['ros_buildfarm']['jenkins']['plugins']` hash to versions compatible with Jenkins 2.555.3, in two stages (see Considerations below).
- (Companion change, separate `jenkins` cookbook PR https://github.com/ros-infrastructure/cookbook-jenkins/pull/16) `cookbooks/jenkins/recipes/jenkins.rb`: add `ignore_failure true` to the initial `service 'jenkins'` start, so a core/plugin bump that leaves Jenkins unable to boot doesn't abort the Chef run before it ever reaches the plugin-install recipe that would fix it.

### Plugin migration

Plugins were migrated in the sequence recommended for a multi-LTS jump, bump the plugin set once on the old core, then bump core, then resolve whatever the new core/JCasC still rejects, then take a final pass at anything newly unlocked:

1. Plugin bump on 2.492.2: updated all non-blocking plugins to their latest versions still compatible with the running 2.492.2 core (see [`2555_minimal_plugins.rb`](https://github.com/user-attachments/files/30359295/plugins.rb.txt) for that snapshot).
2. Moved to Jenkins 2.555.3.
3. Dependency conflict resolution: the first boot after the core bump crash-looped:
   - `matrix-auth` (already pinned to a version requiring newer `cloudbees-folder`/`configuration-as-code`) threw `CannotResolveClassException`, which broke JCasC's ability to parse the authorization strategy and aborted boot.
   - Bumping `cloudbees-folder`/`configuration-as-code` then hit a second wall: `jenkins-plugin-manager` refused to write the plugin set because `configuration-as-code`'s newer release itself transitively requires newer `prism-api`, `bootstrap5-api`, `font-awesome-api`, `plugin-util-api`, `workflow-api`, and `workflow-support` than were pinned.
   - Separately, this surfaced a real ordering bug in the shared `jenkins` cookbook: `jenkins::jenkins` force-starts the service before `ros_buildfarm::plugins` ever runs, so a core/plugin incompatibility permanently deadlocks every future Chef run (Jenkins won't boot until plugins are fixed; plugins can't be fixed because Chef aborts before reaching that recipe). Fixed with `ignore_failure` on that first start, since nothing between it and the later `ros_buildfarm::jenkins` restart depends on Jenkins actually being live.
4. Final pass: after core+plugins booted cleanly, checked the remaining plugins the Jenkins plugin-manager UI flagged as having an "incompatible" newer version available (`ssh-slaves`, `groovy-postbuild`, `badge`) and confirmed via upstream changelogs and a grep of the `ros_buildfarm` job-template repo that none of the removed/changed APIs (`addHtmlBadge`, `addShortText`, `createSummary`, `manager.createSummary(...)`, SSH-launched agent nodes) are actually used anywhere in our generated job configs, before bumping them to `3.605.v7a_3387b_97a_b_6`, `303.v5fe3da_6233f0`, and `3.1097.v868116049892` respectively.

Follow-ups outside this repo:
- `ros_buildfarm` Python job-template repo hardcodes `plugin="groovy-postbuild@@267.va_df06de9fa_fa_"` in `publisher_groovy-postbuild.xml.em`, should be bumped to `303.v5fe3da_6233f0` to match, or JCasC/job-config diffing may flag it.

### Changelogs (ClaudeCode Sonnet 5 Looked for them)

**Jenkins core, 2.492.2 → 2.555.3** (verified against jenkins.io upgrade guides / changelog, not just release announcements):
- **2.492.x**: `agentProtocols` removed from JCasC global config (already handled in this cookbook prior to this PR); JCasC begins hard-failing startup on deprecated fields by default (can be softened to a warning via `deprecated: warn`).
- **2.516.x**: `myViewsTabBar` JCasC field deprecated; DB passwords capped at 72 bytes (bcrypt limit); `jbcrypt` removed from core (Active Directory plugin users need ≥2.40); cookies gain `SameSite=Lax` by default.
- **2.555.1**: **Java 21 or Java 25 required on both controller and agent JVMs — Java 17 support dropped entirely.** Default crumb issuer no longer uses client IP (`excludeClientIPFromCrumb` removed from JCasC).
- **2.555.2**: Windows MSI installer now signed by "LF Open Source, LLC" — SmartScreen will flag it as unrecognized on first install (not applicable to this Linux-only deployment).
- **2.555.3**: Fixes a regression (introduced in 2.523) that could lose active builds when a job configuration was reloaded; fixes an agent-offline-status-not-persisting regression (from 2.482) under JCasC.

**Plugins** (see Plugin migration above for the reasoning; changelog specifics):
- `matrix-auth` 3.2.3+ requires `cloudbees-folder` ≥ 6.1026.ve06dfa_cf31c3 and `configuration-as-code` ≥ 1963.v24e046127a_3f (per its POM).
- `configuration-as-code` 1963+ transitively requires the newer `prism-api`/`bootstrap5-api`/`font-awesome-api`/`plugin-util-api`/`workflow-api`/`workflow-support` versions now pinned.
- `ssh-slaves` 3.x (PR `jenkinsci/ssh-slaves-plugin#541`): removed automatic Java path detection (`JavaProvider`) for SSH-launched agents — must be explicit now.
- `badge` 3.x: removes pipeline steps deprecated since Badge 2.0 (`addHtmlBadge`, `addShortText`, `createSummary`, `summary.appendText`) — replacements are `addBadge`, `addBadge`, `addSummary`, `summary.setText(...)`. Confirmed unused in our generated job configs.
- `groovy-postbuild` 303: its own maintainers acknowledge (`jenkinsci/groovy-postbuild-plugin#190`) that Badge 3.x breaks the freestyle `manager.createSummary(...)` API used from Groovy Postbuild scripts, and the real fix (#191, "Support badge-plugin 3.x") is still open/unmerged upstream. Confirmed no job template in this repo calls `manager.createSummary(...)`, so this doesn't affect us today — but there's no upstream fix to fall back on if that changes.


.github/workflows/main.yml
claraberendsen: I'm not sure we need to run CI for these different chef versions, but it's not a blocker for this PR.

claraberendsen (approved):

attributes/jenkins.rb
```diff
@@ -22,8 +22,7 @@
 
 # Last version supporting Java 8
 #default['jenkins']['master']['version'] = '2.346.1'
```
claraberendsen:
```suggestion

```

Let's cleanup some of these comments.

PR #184, opened 2026-07-31
Double the agent disk space thresholds

cottsay:
We're getting a LOT of disk space issues now that our agents are living longer. When they fill up, they should be taken out of circulation and NOT just sit there and fail builds.

https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/demos/node-monitors/README.md


Crola1702 (approved):
Are we expecting manual intervention for agent decomission? Or is it managed by Jenkins somehow? 🤔

PR #186, opened 2026-08-13
Add anubis support

claraberendsen:
This PR adds Anubis support to the jenkins manager behind a feature flag on the cookbook attributes
https://github.com/ros-infrastructure/cookbook-ros-buildfarm/blob/5732ae35d5b40fb85fc7b4d7aee9a4a0d515d59c/attributes/jenkins.rb#L29-L31

### Implementation notes 
- Anubis is expected to be installed on the system, this cookbook does not install or configures it. 
- Furthermore Anubis is expected to be running when the jenkins recipe runs. 
- Anubis is expected to be running at the value defined at `node['anubis']['socket']'
- A guard will fail the recipe if the attribute is set but Anubis is not running. 
- The test on this cookbook don't exercise/validate the Anubis proxy it only checks that nginx correctly forwards the expected endpoints to anubis. 

Two extra nginx improvements are bundled in ported from config at build.o.o:

https://github.com/ros-infrastructure/cookbook-ros-buildfarm/blob/cc22afe94c1741a0606dae06699cad422743073e/templates/nginx/jenkins-webproxy.ssl.conf.erb#L73-L85

### DIFF
--- a/templates/nginx/jenkins-webproxy.ssl.conf.erb
+++ b/templates/nginx/jenkins-webproxy.ssl.conf.erb
@@ -7,6 +7,17 @@ upstream jenkins {
 	server 127.0.0.1:8080 fail_timeout=0;
 }
 
+# Configure Anubis Upstream
+# It's expected to be running as anubis@jenkins systemd unit
+<% if (node['jenkins']['anubis']) %>
+upstream anubis {
+	server unix:/run/anubis/jenkins/instance.sock fail_timeout=0;
+  # Fallback to serving through nginx to avoid downtime if anubis is down
+  server 127.0.0.1:8080 backup;
+}
+<% end -%>
+
+
 # HTTP server is used just for ACME challenge and allowing build status icon
 # links that are already in the wild to be accepted. Otherwise redirect to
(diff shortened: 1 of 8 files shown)


templates/nginx/jenkins-webproxy.ssl.conf.erb
```diff
@@ -59,6 +70,68 @@ server {
...
+  # see https://github.com/osrf/chef-osrf/issues/367
+  location ~* "/descriptorByName/org\.jenkinsci\.plugins\.ghprb\." {
+      return 403;
+      access_log /var/log/nginx/ghprb_blocked.log;
+  }
+  
+# API endpoints help bots identify how to query the buildfarm
+  # There is no usual usage of them by the dev team so we should disable them
+  location ~* /api/(xml|json|python|schema)(/|$) {
+      return 503;
```
Crola1702: I consider that a better approach that we return `429` or just `403` instead of a `500` code, specially because `500` codes may trigger in bots just a "wait, and try again later", while a `400` might just say: "okay we'll just drop this"

